PR #2, opened 2015-07-21
added checks for compiler flags

PeterMitrano:
Stole this trick from gazebo, basically checks if a flag exists by test-compiling it, and adds it if it works. Be careful though--/Wall on windows is a nightmare...

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,18 @@
 cmake_minimum_required(VERSION 2.8)
 project(ntcore)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++1y -Wformat=2 -Wall -Wextra -Werror -pedantic -Wno-unused-parameter")
+include(CheckCXXCompilerFlag)
+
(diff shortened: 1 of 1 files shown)


PeterJohnson:
Hmm, would it be better just to do if(MSVC) and define MSVC-specific flags?

PeterMitrano:
It might be, but I happened to have this open from gazebo so I thought I'd
offer.

PeterJohnson:
The current flags work on both GCC and clang.  What flags are you using on MSVC?

PeterMitrano:
Updated with MSVC specific flags. the C++ 11 that it has is enabled default, and Wall/Wextra are a way too verbose for my tastes.

PeterJohnson:
Merged modified and squashed version as 9f728b8

PeterMitrano:
Just tried compiling WPILib Sim stuff with vc 2015...not immediately successful. I know it might be annoying, but can we settle for 2013? I'm worried it might require rebuilding gazebo and it's dependencies with msvc 2015.

PR #3, opened 2015-08-15
Gradle changes

333fred:
This adds build support for building with gradle on Linux (x86/x86_64), Windows (x86/x86_64) using MSVC, and ARM. To build on Windows, you must have VS 2013 installed. Gradle seems to not be doing something right with the include paths for the Win10 sdk yet, which is why I've set the VS install path manually. MSVC also appeared to have issues with std::min and std::max in StringRef.h. I'm not at all sure why that's happening, but removing the std:: prefixes and including them manually seemed to fix it. I also added the pragma for linking in ws2_32.lib on Windows, as that's where all the Winsock2 functions are defined.
When this is merged into WPILib proper, we should not be defining the ARM toolchain in this gradle file. That's taken care of in build.gradle in the root of the allwpilib repository, and I've copied that section into here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,35 @@
+### C++ ###
+# Compiled Object files
+*.slo
+*.lo
+*.o
+*.obj
+
+# Precompiled Headers
+*.gch
+*.pch
+
+# Compiled Dynamic libraries
+*.so
+*.dylib
+*.dll
+
+# Compiled Static libraries
+*.lai
+*.la
+*.a
+*.lib
+
+# Executables
+*.exe
+*.out
+*.app
+build/
+.gradle/
+
+.vs/
+*.def
+*.opensdf
+*.vcxproj
+*.vcxproj.user
+*.sdf
--- a/build.gradle
+++ b/build.gradle
@@ -1,9 +1,97 @@
+import org.apache.tools.ant.taskdefs.condition.Os
+
 apply plugin: 'cpp'
+apply plugin: 'visual-studio'
 
 model {
+    toolChains {
+        gcc(Gcc) {
+            target("arm"){
+                // We use a custom-built cross compiler with the prefix arm-frc-linux-gnueabi-<util name>
+                // If this ever changes, the prefix will need to be changed here
+                def compilerPrefix = 'arm-frc-linux-gnueabi-'
+                cppCompiler.executable = compilerPrefix + cppCompiler.executable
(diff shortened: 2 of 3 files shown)


PeterJohnson:
The reason for the min/max issue is that the windows SDK conflicts with the C++ standard.  You need to define NOMINMAX to keep the windows ones from getting defined; I do this on the compiler command line in cmake (see CMakeLists.txt line 5).  The other thing worth defining is _SCL_SECURE_NO_WARNINGS to avoid a bunch of spurious warnings.  Also, do the gradle build files use ntcore.def when linking the DLL?  This forces the entry points to have consistent ordinals.  I'm not a huge fan of pragmas to force linking, but it's probably ok in this case.

333fred:
I've updated the build script with those flags, reverted the changes to stringref, and added the defs flag. The pragma linking is the traditional way to this for Windows programs, and is the way Microsoft recommends doing it, so I'd prefer to go that route, although I certainly understand the want to keep all build-related settings actually in the build files. At this point, the only thing left to add is support for the tests, which will need some stuff with gmock.

PeterJohnson:
Could you squash the commit so it's a cleaner merge?

333fred:
Squashed. I'm currently working getting gtest and gmock working, and I could use a second opinion. With Gradle, as with CMake, you have to have the library yourself, and can't just use dependency resolution as you can with Java. However, that's significantly harder to do with Gradle than it is with CMake. In fact, the official gtest example from Gradle has them including the libraries pre-built with the repository for all relevant platforms, and selecting which one to use based on current OS. Thoughts on this?

PeterJohnson:
Pre-built binaries sounds really ugly.  The cmake build setup I have currently actually downloads the gtest/gmock source and compiles it.  We don't need to be that fancy; I'd be okay with including a copy of the gmock source code but would really prefer not putting binaries in the source tree.  Hmm, I wonder if anyone's hosting a gmock repo that git externals could grab.  When this gets moved into the wpilib tree there's more copies of gmock already there... it'd be nice to consolidate as long as building ntcore independently still works (within a larger checkout)

PeterJohnson:
Should I go ahead and merge this and you'll make the gtest/gmock changes a separate commit?

333fred:
I would say yes. I had been hoping to get it done over the weekend, but I wasn't quite able to get it done. Hopefully I'll have it finished sometime this week.

PR #4, opened 2015-08-20
Replaced time.h with std::chrono

calcmogul:
This implementation returns the same values as the previous one on both a Linux machine and the roboRIO.


PeterJohnson:
The previous code used REALTIME for the initial base time read, but MONOTONIC for subsequent reads, which means we should probably use steady_clock instead of high_precision_clock?

calcmogul:
The original was using MONOTONIC to add the time elapsed since the initial time. I've updated the commit to do that instead.

PR #5, opened 2015-08-20
Starting to add Getters, Setters, and Allocators to the C interface

ThadHouse:
Because of the union in the NT_Value structure, some languages would
have a difficult time with the interop. This commit adds individual
getters and setters to the C interface to make interop easier.

In addition, some languages cannot allocate native memory in the same
heap that the C code would use. This adds allocation functions to make
sure that all our memory is allocated and can be freed properly from the
same heap.

The reason the getters are not individual get functions are because that
would require the calling code to first get the type, and then call the
specific getter for that type. Doing combined get functions causes
interop code to only have to cross interop boundaries once to get the
value instead of twice, for a light performance increase. However, this
could be changed without much work.

This pull request is not complete. I plan on adding comments, and moving the functions to better locations in the file. However I wanted to get opinions on how the interface looks, and on if the actual C code is even properly written. This is the most C code I've ever written, and want to make sure I'm not doing it the completely wrong way.

### DIFF
--- a/src/ntcore_c.cpp
+++ b/src/ntcore_c.cpp
@@ -76,6 +76,10 @@ static void DisposeConnectionInfo(NT_ConnectionInfo *info) {
   std::free(info->remote_name);
 }
 
+static void DisposeEntryInfo(NT_EntryInfo *info) {
+  std::free(info->name.str);
+}
+
 static RpcParamDef ConvertFromC(const NT_RpcParamDef& in) {
   RpcParamDef out;
   out.name = ConvertFromC(in.name);
@@ -409,8 +413,19 @@ void NT_InitString(NT_String *str) {
   str->len = 0;
 }
 
+enum NT_Type NT_GetType(const char *name, size_t name_len) {
+  auto v = nt::GetEntryValue(StringRef(name, name_len));
+  if (!v) return NT_Type::NT_UNASSIGNED;
+  return v->type();
+}
+
 void NT_DisposeConnectionInfoArray(NT_ConnectionInfo *arr, size_t count) {
   for (size_t i = 0; i < count; i++) DisposeConnectionInfo(&arr[i]);
+  std::free(arr);
+}
+
+void NT_DisposeEntryInfoArray(NT_EntryInfo *arr, size_t count){
+  for (size_t i = 0; i < count; i++) DisposeEntryInfo(&arr[i]);
   std::free(arr);
 }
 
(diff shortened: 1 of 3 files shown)


PeterJohnson:
Would it be cleaner (and almost certainly a little more efficient) to simply provide type-specific SetEntryValue/SetEntryTypeValue functions rather than utility functions to create/destroy NT_Value*'s that are basically only used to pass to SetEntryValue/SetEntryTypeValue?  To set a single value this way, you have to call 3 native functions, versus 1 with type-specific SetEntryValue/SetEntryTypeValue.  To cut the number of unique functions in half, we could merge SetEntryValue and SetEntryTypeValue into a single function with a "force" parameter.  The higher level language could easily separate them back out.

ThadHouse:
That's better for sure. I wasn't even thinking that way. Ill change it.

For the RPC functions that use NT_Values, what exactly are those used for? Since those use NT_Values, those might have to have some wrapper code as well, unless the can just use straight values returned by the NT_GetEntryValue function.

PeterJohnson:
The RPC functions are designed as helper functions; the RPC itself simply provides the RPC input/output data as an encoded string, so you can either use the RPC helper functions to decode it or decode it yourself in the higher level language (likely better).  Note callbacks use NT_Values as well.  This is one area where providing type-specific accessor functions into an existing NT_Value\* probably makes sense, despite the extra function call.

ThadHouse:
Is RPC something new to NT 3.0? I guess I just don't understand what  it actually is, and I don't remember seeing anything called RPC when I was porting over the old network tables.

NT_GetNTValueData was written for the callbacks mainly. I was thinking about using type specific accessors here, but doing that would have required 2 native function calls, whereas dumping everything into a void\* and having the higher level language filter that only required 1 native function call.

PeterJohnson:
Ok, NT_GetNTValueData should be sufficient.  I'm not a huge fan of void\* but it seems reasonable here.

Yes, RPC is new to NT 3.0.  The server can define remote procedure calls (with arbitrary input and output parameters) by defining specially-typed NT values, and the client can call them.  This is a cleaner and much more flexible way of handing GUI-ish things like dashboard buttons via NT (rather than monitoring a value and writing to it).

333fred:
I'm really not a fan of the void\* here either, especially not without a size parameter on it. Would it be unreasonable to add an input size parameter? In NT_GetNTValueData, just add a size_t parameter, and in NT_GetNTValueDataArr, have the size_t be the max size on input and returned size on output?
Also, for the NT_GetNTValueDataArr function, you're not setting the value of size. You're setting the pointer to point to a size, which isn't reflected in the calling function (unless I'm just hallucinating).

333fred:
Or am I misunderstanding the data parameter? If it's not actually a copy, then it will also have the same issue, as you're setting the value inside the function that changes nothing outside the function.

ThadHouse:
I screwed up NT_GetNTValueDataArr, and actually have removed it. I just haven't commited the changes. 

The data parameter I want to return a pointer to whatever value is stored in data. The calling side can then use the returned enum to figure out what the size of the data is. Individual Getters could be added, but that would end up requiring 2 calls to native code for every time a value needs to be grabbed. Doing a void\* ends up only requiring 1 call, and then the caller can use the type enum to cast the void\* to its actual value.

333fred:
Alright, then data needs to be a `void**`, not a `void*`. Currently, you're setting

``` C
data = value->data->type.arr;
```

where data is a `void*`, what you want to be doing is

``` C
*data = value->data->type.arr;
```

where data is a `void**`.

ThadHouse:
Why does it have to be a void_**? Can't void_\* be directly cast into an int*? Testing last night with the way is written actually worked correctly.

Sorry I don't really write C code, so working with pointers is a little difficult sometimes.

333fred:
Sorry, the `void**` was for NT_GetNTValueDataArr, not NT_GetNTValueData. Sure, `void*` can be directly cast into an `int*`. But the problem is that you're not changing what the caller sees. As the caller, I'll have an integer, say `int i;`. When I call NT_GetNTValueData as currently written, I'll pass a pointer to i, so `NT_GetNTValueData(..., &i, ...);`. This pointer is passed _by value_, not by reference. So data is a copy of the pointer to i, and you modify your copy of the pointer to point to the raw value, but you do not set the value of i, which is what you intended. You want to do:

``` C
// Sets what data points to, which is i. This is reflected in the caller with a change in i.
*data = value->data->type.raw;
```

not

``` C
// Sets data, which is a local variable and is not reflected in the caller. Further, you no longer
// know where integer was located, since we've overwritten the pointer.
data = &value->data->type.raw;
```

Pointers take some getting used to. Everyone struggles with them, so don't worry too much.

ThadHouse:
Ah that makes sense. Ill work on fixing it this afternoon. I think I figured out a way to deal with void**'s from C#, which was what I was worried about. But it should be fairly easy to do.

ThadHouse:
I think I got all of the issues fixed. The only function I'm not too sure about is NT_GetDataFromNTValue. The array cases should be write based off of Fredric's comments, but I'm not quite sure about the single value cases. Are my pointers and references right there?

src/ntcore_c.cpp
```diff
@@ -436,3 +436,174 @@ void NT_DisposeRpcCallInfo(NT_RpcCallInfo *call_info) {
...
+void NT_GetDataFromNTValue(struct NT_Value *value,
+	void *data, void **data_arr, size_t *size, enum NT_Type *type,
+	unsigned long long *last_change)
+{
+	*type = value->type;
+	*last_change = value->last_change;
+	switch (value->type)
+	{
+	case NT_Type::NT_BOOLEAN:
+		data = &value->data.v_boolean;
```
333fred: Almost. You want to be doing

``` c
*((int*)data) = value->data.v_boolean;
```

and cast to the appropriate types for the other ones. Also, we should be setting size to the size of what we are writing, so in this case

``` c
*size = sizeof(value->data.v_boolean);
```

src/ntcore_c.cpp
```diff
@@ -436,3 +436,174 @@ void NT_DisposeRpcCallInfo(NT_RpcCallInfo *call_info) {
...
+	case NT_Type::NT_DOUBLE_ARRAY:
+		*data_arr = value->data.arr_double.arr;
+		*size = value->data.arr_double.size;
+		break;
+	case NT_Type::NT_STRING_ARRAY:
+		*data_arr = value->data.arr_string.arr;
+		*size = value->data.arr_string.size;
+		break;
+	default:
+		data = nullptr;
```
333fred: A couple of things. First, we don't want to change data here. What we will want to do is set *data_arr to be NULL.

333fred:
Another thing. I don't think that I like having the single types and array types together like this. If we do, we should be doing null checks on the data and data_arr parameters to make sure they're not null pointers, but only when appropriate. For example, if the data type is a single data type, it's conceivable that the user could accidentally mess up the parameter order and pass NULL to the data parameter instead of the data_arr parameter. @PeterJohnson, thoughts?

ThadHouse:
After doing some testing, that code is even having trouble on the C# side as well. So I think I'll change it to seperate getters.

The reason they were 1 combined function is that NT_Value structs to my C# code can only be opaque pointers, since I can't create a managed version of the struct. So because of that, doing it all in a single function didn't require me to first get the type, and then call its specific getter, which would be 2 native calls. 

Its going to require 15 interface functions instead of 2, but I think the code would be much cleaner that way.

ThadHouse:
I switched everything to individual Getters. Because of this I had to add Free functions for the arrays. But this should work much better and cleaner.

src/ntcore_c.cpp
```diff
@@ -436,3 +436,257 @@ void NT_DisposeRpcCallInfo(NT_RpcCallInfo *call_info) {
...
+{
+	std::free(v_double);
+}
+void NT_FreeBooleanArr(int *v_boolean)
+{
+	std::free(v_boolean);
+}
+void NT_FreeStringArr(struct NT_String *v_string, size_t arr_size)
+{
+	for (size_t i = 0; i < arr_size; i++)
```
333fred: We don't allocate strings as part of AllocateNTStringArr, so doing this here isn't expected behavior. The user has to allocate the NTString themselves, so it should be their responsibility to deallocate them. All we should do here is free the array.

src/ntcore_c.cpp
```diff
@@ -436,3 +436,257 @@ void NT_DisposeRpcCallInfo(NT_RpcCallInfo *call_info) {
...
+	return ret;
+}
+NT_String *NT_GetEntryStringArr(const char* name, size_t name_len, unsigned long long *last_change, size_t *arr_size)
+{
+	NT_Value *value = static_cast<NT_Value*>
+		(std::malloc(sizeof(struct NT_Value)));
+	NT_GetEntryValue(name, name_len, value);
+	NT_String *ret = NT_GetEntryStringArrFromValue(value, last_change, arr_size);
+	return ret;
+}
```
333fred: Make sure you have a newline at the end of the file, that's a POSIX standard we should try to conform to.

src/ntcore_c.cpp
ThadHouse: I did this to match NT_DisposeValue. Should it be deallocating them there as well?

333fred: Hmm. I'll punt this to @PeterJohnson.

PeterJohnson: NT_DisposeValue should free any internals because the code allocates them in NT_GetEntryValue, the most common use case.  In general I like freeing internals, although it does risk double-frees if the user frees it themselves without setting the pointer to null afterwards.

One thing that should be cleaned up in the C API is that only DisposeConnectionInfoArray frees the passed pointer while all other functions do not, which is confusing (while the code allocates the ConnectionInfo array and thus should be responsible for freeing it, the function probably shouldn't use the Dispose\* name).

ThadHouse:
Oh. I didn't even know those existed. That should be much better.

And that's why I was returning the created NT_Value struct. The caller could then call DisposeValue to dispose of it. But a way to do it without malloc'ing will work much better.

ThadHouse:
Is there any reason there is no Value::MakeStringArray that takes an ArrayRef of NT_Strings? Or is there a method to convert an NT_String array to a std:string array?

ThadHouse:
I used the code used to convert the value to get the NT_String array converted. The only mallocs remaining are the ones to allocate arrays.

I do have a question about the allocations though. When you set an array to a value in the table, can you then free the memory for that array? Same thing if you write an NT_String to the table.

PeterJohnson:
Yes, the Make\* functions make a copy of the provided data, so you should free the memory for the array/NT_String after passing it.  We can certainly add a Value::MakeStringArray that takes an ArrayRef of NT_String; that would avoid the extra array copy you're incurring now.

One formatting thing: we use two spaces for indentation, it looks like you're using tabs?

ThadHouse:
Good to know about the allocation. I'll make sure to remember that when doing the implementation. I'm thinking that since the SetEntryValue function does not free the arrays, and requires them to be done by the caller these new functions should act the same.

I looked into making a Value::MakeStringArray, but the Value struct uses std::strings, so adding a Value::MakeStringArray would require making the copy anyway it seems. The SetEntryValue function also does a copy from NT_String to std::string, so doing that in the individual setter isn't any more work.

As for formatting, I wanted to make sure everything was right since I don't really know C, and then I was going to go through and properly format, then squash the commits together.

ThadHouse:
For the GetEntryDouble and GetEntryBoolean functions, what do you think is the best way to return that the value is not set? I was currently returning 0, but I realized that could be an actual number, and not good for error checking. I could either return something like the minimum value for the number, or add another parameter that returns a status of if it exists. Returning minimum value would be cleaner, but it makes that 1 number unusable. Using a status is probably safer, but requires creating an extra variable every time on the calling side. Which option do you guys think we should go with?

333fred:
This is when I want Rust enums, they work perfectly for this type of situation, :). I think the best option is go with a status. I don't really view that as an extra variable, especially since there's no real value that we can use a sentinel for doubles (as all doubles are valid NT values, even NaN), and I just morally hate the idea of a tri-state Boolean.

ThadHouse:
Should I use the same signatures on all functions? Technically only GetDouble and GetBoolean need it, since an NT_String returned with a nullptr for its array should be seen as not set, and returning nullptr for the pointer returns is the right way to do it anyway.

ThadHouse:
Squashed all the commits, and fixed the formatting.

I also added an NT_ContainsKey method. If only using the external C interface, there was no way to check if the key existed without allocating an entire NT_Value struct that wouldn't even get set. The Contains Key method checks if a key exists without having to allocate an entire NT_Value struct. I added this to the main utility functions, because this seems like it would be useful even from Java.

PeterJohnson:
Thanks.  Good point on Contains, although a GetType function is likely even more useful, as it both provides existence and you can switch off of it.  Originally GetEntryInfo was intended for this, but it morphed into something much more complicated (more like an enumerate than a get type).

ThadHouse:
I was actually just thinking that as well. I'll at it in in a bit.

ThadHouse:
Everything should be fully implemented, and the code should be formatted correctly.

I have the C# side written too, but I need to create a VM with VS 2013 on it. It doesn't seem to want to build with VS 2015. The native library is nice though, because the code went from 2000 lines and many files to 300 lines with 4 files.

When the EntryListener callback is called, does the receiver have to free the key string? Or is that done by the caller when the callback returns?

333fred:
What doesn't build with VS 2015? Are you using Gradle or cmake?

ThadHouse:
Hmm. I was getting a linker error yesterday. It did just seem to work though.... 

Edit: Ran the wrong build. Still not building properly. Using Gradle.

333fred:
I know that, for Gradle, I detect whether VS 2015 is installed and default to 2013 instead, as there are linking issues with the Gradle visualCpp toolchain currently. I don't know about cmake though.

ThadHouse:
What went wrong:
Execution failed for task :linkX64NtcoreSharedLibrary.
 A build operation failed.
      Linker failed while linking ntcore.dll.
  See the complete log at: file:///C:/Users/thad/Documents/GitHub/ntcore/build/tmp/linkX64NtcoreSharedLibrary/output.txt

Error from file.
LINK : fatal error LNK1104: cannot open file 'libcpmt.lib'

This is the error I am getting from gradle. 

I don't have VS 2013 installed on my machine. I want to get a VM up with is, but don't have time today.

ThadHouse:
```
See file:///C:/Users/thad/Documents/GitHub/ntcore/build/tmp/compileArmNtcoreSharedLibraryNtcoreCpp/output.txt for all output for compileArmNtcoreSharedLibraryNtcoreCpp.
compiling ntcore_c.cpp failed.
In file included from C:\Users\thad\Documents\GitHub\ntcore\include/ntcore_cpp.h:20:0,
                 from C:\Users\thad\Documents\GitHub\ntcore\include/ntcore.h:16,
                 from C:\Users\thad\Documents\GitHub\ntcore\src\ntcore_c.cpp:8:
C:\Users\thad\Documents\GitHub\ntcore\include/nt_Value.h: In instantiation of 'static std::shared_ptr<nt::Value> nt::Value::MakeString(T&&) [with T = NT_String&]':
C:\Users\thad\Documents\GitHub\ntcore\src\ntcore_c.cpp:531:80:   required from here
C:\Users\thad\Documents\GitHub\ntcore\include/nt_Value.h:114:19: error: no match for 'operator=' (operand types are 'std::string {aka std::basic_string<char>}' and 'std::remove_reference<NT_String&>::type {aka NT_String}')
     val->m_string = std::move(value);
                   ^
C:\Users\thad\Documents\GitHub\ntcore\include/nt_Value.h:114:19: note: candidates are:
In file included from c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\string:52:0,
                 from c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\stdexcept:39,
                 from c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\array:38,
                 from c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\tuple:39,
                 from c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\functional:55,
                 from C:\Users\thad\Documents\GitHub\ntcore\include/ntcore_cpp.h:12,
                 from C:\Users\thad\Documents\GitHub\ntcore\include/ntcore.h:16,
                 from C:\Users\thad\Documents\GitHub\ntcore\src\ntcore_c.cpp:8:
c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\bits\basic_string.h:554:7: note: std::basic_string<_CharT, _Traits, _Alloc>& std::basic_string<_CharT, _Traits, _Alloc>::operator=(const std::basic_string<_CharT, _Traits, _Alloc>&) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]
       operator=(const basic_string& __str) 
       ^
c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\bits\basic_string.h:554:7: note:   no known conversion for argument 1 from 'std::remove_reference<NT_String&>::type {aka NT_String}' to 'const std::basic_string<char>&'
c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\bits\basic_string.h:562:7: note: std::basic_string<_CharT, _Traits, _Alloc>& std::basic_string<_CharT, _Traits, _Alloc>::operator=(const _CharT*) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]
       operator=(const _CharT* __s) 
       ^
c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\bits\basic_string.h:562:7: note:   no known conversion for argument 1 from 'std::remove_reference<NT_String&>::type {aka NT_String}' to 'const char*'
c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\bits\basic_string.h:573:7: note: std::basic_string<_CharT, _Traits, _Alloc>& std::basic_string<_CharT, _Traits, _Alloc>::operator=(_CharT) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]
       operator=(_CharT __c) 
       ^
c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\bits\basic_string.h:573:7: note:   no known conversion for argument 1 from 'std::remove_reference<NT_String&>::type {aka NT_String}' to 'char'
c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\bits\basic_string.h:589:7: note: std::basic_string<_CharT, _Traits, _Alloc>& std::basic_string<_CharT, _Traits, _Alloc>::operator=(std::basic_string<_CharT, _Traits, _Alloc>&&) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]
       operator=(basic_string&& __str)
       ^
c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\bits\basic_string.h:589:7: note:   no known conversion for argument 1 from 'std::remove_reference<NT_String&>::type {aka NT_String}' to 'std::basic_string<char>&&'
c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\bits\basic_string.h:601:7: note: std::basic_string<_CharT, _Traits, _Alloc>& std::basic_string<_CharT, _Traits, _Alloc>::operator=(std::initializer_list<_Tp>) [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]
       operator=(initializer_list<_CharT> __l)
       ^
c:\frc\arm-frc-linux-gnueabi\include\c++\4.9.1\bits\basic_string.h:601:7: note:   no known conversion for argument 1 from 'std::remove_reference<NT_String&>::type {aka NT_String}' to 'std::initializer_list<char>'

Finished compileArmNtcoreSharedLibraryNtcoreCpp, see full log file:///C:/Users/thad/Documents/GitHub/ntcore/build/tmp/compileArmNtcoreSharedLibraryNtcoreCpp/output.txt. 
```

This happens when trying to build without NT_SetEntryString and NT_SetEntryRaw commented out. I think is having trouble trying to make a copy of the string. I don't know enough about the behind the scenes stuff to know why this error would be happening.

ThadHouse:
Got the returning struct thing fixed, and got string returning to work. Just need to build the VS 2013 system to test and this should be basically done except for documentation.

PeterJohnson:
Great!  Your current approach to array and non-array getters looks good to me.  I'm still on the fence about having a "force" parameter versus just having separate functions (I can see arguments both ways).  One remaining item: can you add the new functions to the end of ntcore.def so they have fixed entry point numbers?

Regarding the gradle build issue you ran into, can you file a separate bug for that if you're still having issues so we don't lose track of it?

ThadHouse:
I added the new functions to ntcore.def.

For the connection and entry listener callbacks, which side has to dispose of the key string? And for the getters and setters is it ok for the caller to create the key string, run the function, and then dispose of the key string? 

I'll create a new issue for the gradle build. I'm assuming it has something to do with only having VS 2015 installed on my machine, with gradle assuming VS 2013.

PeterJohnson:
The callbacks get const char\* strings / const\* values; the library owns them and the callee should NOT attempt to free them.  The key strings for getter/setters are const char\* as well.  The library always makes a copy of the provided string (or only uses it temporarily for the duration of the call) so the caller is responsible for freeing the string if it allocates memory for it.  Note plain const char\* / length are taken for the key strings instead of NT_String.  This is intentional as most languages have some way of getting a raw char\* from an internal language string, avoiding making a copy.

ThadHouse:
Ha most languages. .Net on windows marshals as either ANSI or UTF-16, whereas mono marshals as  UTF-8... So I'm having to manually marshal everything as UTF-8, and copy it to a pointer. I might try and see if there is a way around this copy, but looking online I'm not seeing any way to do this. 

Edit Figured out I can pass those as byte[]'s instead, and it should just pin them instead of copying. Still have to manually convert to UTF-8, but 1 copy is better then 2.

ThadHouse:
Should I squash the rest of the commits together? I was able to test most things today and everything seemed to work.

PeterJohnson:
Yes, go ahead.  I'll review it again after the squash.

ThadHouse:
OK squashed into 1 single commit.

ThadHouse:
I don't think I did the rebase right... It added every commit added since I branched to the pull request. How do I fix that?

PeterJohnson:
Try the steps here: http://stackoverflow.com/questions/17182624/contributing-to-project-on-github-how-to-rebase-my-pull-request-on-top-of-mast

ThadHouse:
I guess I had to add upstream. I need to learn how to use cmd git better. I've been using the GitHub app for a while, and its gotten me into bad habits. 

Its fixed now.

ThadHouse:
Should I make all of values given to the setters const? If the code makes deep copies of everything it writes, that seems like a good idea.

PeterJohnson:
Yes, please do so.

PeterJohnson:
Can you rebase and fix the conflicts in ntcore.def?  Also, I would prefer that SetEntryString and SetEntryRaw take const struct NT_String\* v_string/v_raw (rather than passing the structs by value) due to possible compiler binary incompatibilities.

ThadHouse:
Fixed.

PeterJohnson:
Sorry, should have been more clear in my second sentence.  Can you make SetEntryString and SetEntryRaw take a pointer to the NT_String structure (NT_String*) rather than passing it as a value (NT_String)?

ThadHouse:
Is there a specific reason why taking a pointer is better then taking a value? Most interop languages store structs on the stack, and passing them by pointers instead of by value requires a lot more work by the interop code to pin that structure and get a reference to it. I'm fine doing it, just want to understand why pointers are better then values.

PeterJohnson:
After investigating a bit more, I think it's okay since the struct is not packed and doesn't have bitfields.  Since you're passing it by value, you don't need the "const", so you can take that back off.

ThadHouse:
Const is removed.

PeterJohnson:
Looking one more time at the passing of a struct thing.. is there a reason why we can't just pass the char\* and length as separate parameters rather than passing a structure?  Some languages like Haskell FFI can't deal with struct parameters.

ThadHouse:
We absolutely could. I could handle that in my code just fine, and in fact it would save a copy. The getters could also return a char\* instead of an NT_String as well. The only reason I didn't was I figured using NT_Value fit the structure of the rest of the code better, but I think passing a char\* would be just fine in this case.

PeterJohnson:
Sounds good, including the getters (of course you'll still need to return the size through a parameter in that case, just like the array getters).  Go ahead and make it a separate pull request.  Thanks!

Just noticed another thing: NT_AllocateNTString returns a struct.  That's a bigger no-no than passing a struct as it's very compiler-specific... is there a reason you can't use a pointer-to-NT_String output parameter for that function?

ThadHouse:
Actually, I might not even need that allocator. The only code that actually used it anymore was Rpc code, and even there it looks like I can ignore it, and just use the native allocator, In fact because everything is consts, the only allocator I still need is NT_AllocateCharArray. So I can remove that completely. I was actually thinking to change that to a NT_AllocateArray function that took a size, and allocated any size, instead of just having a specific allocator. Is that an OK idea?

PeterJohnson:
I prefer the type-specific functions like Char/Double/Boolean for array allocation as you maintain type safety.

ThadHouse:
Ok. However should I delete any I don't need, or leave them in, and just fix NT_String?

PeterJohnson:
I would remove AllocateNTString (because it's not required given the existence of AllocateCharArray) and leave the rest.

PeterJohnson:
Went ahead and made the changes.  I also changed NT_GetFromValue to return copies rather than references.  It's an extra copy but returning references to within the NT_Value is inconsistent with NT_GetEntry and risks double-frees.

ThadHouse:
I just realized if we are returning a char\* that's allocated with malloc, we need a way to free that array afterward. Should I create a new pull request for that, or do you want to do that?

PeterJohnson:
Done.

Issue #13, opened 2015-08-22
No entry points found for Windows DLL's.


ThadHouse:
I was able to get a successful build on a VS 2013 machine, but when trying to open the DLL it says it cannot find the entry point its looking for. Running dumpbin on the compiled DLL shows no functions at all being exported from the DLL. Are there any special gradle flags needed to get the exports to build correctly?

PeterJohnson:
I've always used cmake to build on Windows.  The ntcore.def looks like it's being used in gradle too, but maybe something's broken there.  @333fred, help?

ThadHouse:
The CMake build does work. I did find a typo in the ntcore.def file for the NT_PostRpcResponse function, which was NT_PostRpcRepsonse. I added that fix to the pull request, which fixed my cmake builds, but it did not fix the gradle build.

333fred:
Sorry for the lack of responses guys. I'm currently transferring planes,
I'll be able to look tomorrow. As to msvc 2015, gradle has some issues with
it and doesn't include the right headers, causing compilation issues.

The CMake build does work. I did find a typo in the ntcore.def file for the
NT_PostRpcResponse function, which was NT_PostRpcRepsonse. I added that fix
to the pull request, which fixed my cmake builds, but it did not fix the
gradle build.

333fred:
Thad, I've just pushed a pull request, #17, which should fix the gradle build.

PR #17, opened 2015-08-23
Fixed Gradle build to actually export proper functions

333fred:
I forgot to set the linker to have the .defs file, and set it on just the compiler. This fixes that. It also fixes the incorrect spelling in the ntcore.defs file.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,7 +67,10 @@ model {
                 }
                 eachPlatform {
                     cppCompiler.withArguments { args ->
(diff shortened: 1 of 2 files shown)


PeterJohnson:
Should the /DEF flag be removed from the compiler line? (e.g. does it do anything there?)

333fred:
Should fix #13.

PeterJohnson:
This change also changes the exportedHeaders to include everything in the src directory.  I'm not sure this is desirable?

333fred:
Updated. Those both were from the plane, I forgot to remove them from the final commit.

PR #18, opened 2015-08-23
Added support for google test running

333fred:
This commit adds support for building and running the test suites on both Linux and Windows. To build all possible components and run the tests, run grade build. To just build all test-relevant components, run gradle :test:build. Note that on Windows, two tests are currently failing (they pass on my Linux machine):
[  FAILED  ] StorageTestsPersistent/StorageTestPersistent.SavePersistent/0, where GetParam() = false
[  FAILED  ] StorageTestsPersistent/StorageTestPersistent.SavePersistent/1, where GetParam() = true


333fred:
The important files to look at the new .gradle files. I've created a project for gmock/gtest, and a project for the tests proper.

PeterJohnson:
Do we really need to bring all of these files into the source tree?  For example, there are two copies of gtest.h, one in gmock/fused-src and one in gtest/fused-src.  And we are never going to use the autoconf bits, only cmake or gradle.

333fred:
I've removed everything I feel comfortable with, removing almost 100 files. Let me know if there's anything else you think I can remove.

PeterJohnson:
Merged with some further reductions and cmake integration as well.

Issue #12, opened 2015-08-22
Builds fail with Visual Studio 2015.


ThadHouse:
On my system I only have VS 2015 installed, and I get a linker error when gradle tries to compile the program.

```
See file:///C:/Users/thad/Documents/GitHub/ntcore/build/tmp/linkX64NtcoreSharedLibrary/output.txt for all output for linkX64NtcoreSharedLibrary.
linking ntcore.dll failed.
LINK : fatal error LNK1104: cannot open file 'libcpmt.lib'

Finished linkX64NtcoreSharedLibrary, see full log file:///C:/Users/thad/Documents/GitHub/ntcore/build/tmp/linkX64NtcoreSharedLibrary/output.txt.
```

I noticed that looking in the options.txt file, it puts the VS directory to `"/LIBPATH:C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\lib\amd64"`, which is for VS2013. Is it supposed to be detecting VS 2015 and switching to it?

PeterJohnson:
@333fred, this is the VS 2015 gradle build issue.

333fred:
Gradle currently has issues with VS2015, so this is expected. I actually detect if you have VS 2015 and switch to 2013, I should probably add checking if you only have vs2015 and erroring out.

PeterJohnson:
Check added in 21e21d3b8bf6f99de236d2eb41d895403f75d7ff.

PR #27, opened 2015-10-04
Gradle Update

### LINKED ISSUE #23: Build on Mac
There doesn't seem to be any way to build this project on a mac. Or at least the instructions are lacking. Is there any known way to compile on a Mac?

333fred:
This reworks the .gradle files to be far more clean, readable, and to support the gradle changes that will be coming to WPILib. This makes the following changes:
1. The different things that main build.gradle does have been separated out. Java things are in the java.gradle, the different toolchains for each system are in the respective toolchain.gradle for that platform, the maven publishing routines are in publish.gradle, and the test configuration is now in test.gradle.
2. Gradle now builds for your current platform by default, and not for arm. You can enable arm builds by specifying the command line flag `-PbuildPlatform=arm`. The main wpilib build will do this by default.
3. Gradle will only build tests when building for the native platform.
4. VS2015 is now supported, fixing #12.
5. Mac builds _should_ work now. However, I don't have a mac to test on, so someone with one to test would be good. Hopefully fixes #23.


PeterJohnson:
Might be cleaner if you squash some (all?) of these commits?

333fred:
Rebased.

PeterJohnson:
If I'm reading this correctly, you have to build native and arm with two separate build commands?  Or does enabling arm build both native AND arm?  Is this what we really want (versus having separate arm and native target names that you could build all with one gradle build)?

333fred:
That's correct, there's two separate build commands. 
I do think this is what we want. I can't see a scenario when we want to build both arm and just one platform. The vision I currently have the multi-platform build is having 4 projects, one for each platform, and having Jenkins merge the jar files in the end. Once the changes I'm making to the wpilib build process are merged, everything that is in the wpilib repo will be either simulation, which is current platform and the buildPlatform variable will be set appropriately, or will be arm. And if we're just building ntcore by itself, we're probably building it to run in a current platform program, such as SmartDashboard or OutlineViewer, so arm is unnecessary. Does this make sense?

333fred:
Whoops, hit tab one too few times.

PeterJohnson:
Hrm, well, that works nicely for a buildbot, but locally we quite often build both.  For example, building wpilib currently builds both the local tools and the stuff for the robot.  I agree that in cmake or make projects doing so takes multiple builds because of the need to set cross-compiler settings uniquely for each build configuration, but I thought one of the advantages of gradle was you could build multiple configurations all at once?  If this is the direction we want to take all of wpilib, though, I guess I'm ok with making ntcore match it.

333fred:
My hope is that with the rest of wpilib, we'll separate out the local tools from the allwpilib repo. For example, I plan on moving OutlineViewer to an entirely separate project, and it will get the ntcore dependency like SmartDashboard or SFX. The allwpilib repo should only build things for arm. True, the plugins will still be built by the allwpilib repo, but they will only be installing arm-specific stuff, not any of the local tools.

gmock/build.gradle
```diff
@@ -1,45 +1,68 @@
...
+    toolChains {
+        if (OperatingSystem.current().isWindows()) {
+            visualCpp(VisualCpp) {
+                // Workaround for VS2015 adapted from https://github.com/couchbase/couchbase-lite-java-native/issues/23
+                def VS_2015_INCLUDE_DIR = "C:/Program Files (x86)/Windows Kits/10/Include/10.0.10240.0/ucrt"
+                def VS_2015_LIB_DIR = "C:/Program Files (x86)/Windows Kits/10/Lib/10.0.10240.0/ucrt"
+
+                eachPlatform {
+                    cppCompiler.withArguments { args ->
+                        if (file(VS_2015_INCLUDE_DIR).exists()) {
```
PeterJohnson: This basically tests if VS2015 is installed, not if it's being used to build.  On a system where both VS2013 and VS2015 are installed, will this break attempts to build with VS2013?  Is it possible to actually determine the compiler version from Gradle?

333fred:
As far as I'm aware, there isn't a way to force gradle to build with a specific vs version short of modifying the toolchain. So I don't think this is a concern.

PeterJohnson:
Ok.  The only other thing I've noticed so far is a carryover from the old build files of applying -O0 -g all the time.  At some point we should really make debug builds an option (the code is /significantly/ faster when built with optimizations turned on).

java/java.gradle
```diff
@@ -0,0 +1,140 @@
...
+                    platformSpecificIncludeFlag(file.getPath(), cppCompiler)
+                }
+            }
+
+            dependsOn jniHeadersNetworkTables
+        }
+    }
+}
+
+ext.getPlatformPath = { binary ->
```
PeterJohnson: This doesn't seem to be working on my machine.  The build ntcore.jar has the library in linux/arm-v7 instead of Linux/arm, and similarly for ntcore-linux.jar.

333fred: Peter, I'm confused. Are you talking about the publishing routine?

333fred: Hmm. This is copied directly from the previous version, so I'm not sure why it would be failing. I'll look into this.

333fred: The strings appear to have changed with Gradle version 2.7. Luckily, there is now architecture Booleans we can check directly. This resolves the arm version, but I don't have a Linux machine handy to test on.

ThadHouse: I tried building this on linux, and I get this error. I don't know if its a setup error or not. This was a fresh ubuntu 14.04 LTS build.

```
* Where:
Build file '/home/thad/git/ntcore/build.gradle' line: 114

* What went wrong:
A problem occurred configuring root project 'ntcore'.
> Exception thrown while executing model rule: ComponentModelBasePlugin.Rules#collectBinaries
   > No signature of method: org.gradle.api.internal.file.copy.CopySpecWrapper_Decorated.getPlatformPath() is applicable for argument types: (org.gradle.nativeplatform.internal.DefaultSharedLibraryBinarySpec_Decorated) values: [shared library 'ntcore:x64:sharedLibrary']
```

333fred:
Thad, are you using the gradle wrapper, or a preinstalled gradle binary on your system? The gradle wrapper is using version 2.7, make sure that's what you're using to build with. I just tested on my Arch laptop, I don't get that exception when using the wrapper.

333fred:
I just updated the pull request again, I had an extra debug statement I put in while fixing the getPlatformPath issue that I forgot to remove.

ThadHouse:
I was calling it with ./gradlew build -PskipJava (The -PskipJava was because the system only has Java 7, and is failing trying to build the java library) I don't know enough about linux or gradle to know everything that has to be set up, or how to get Java 8 installed properly.

ThadHouse:
Also updating to the new files doesn't change anything. I'm going to assume its a setup error. But it would be nice to be able to build the linux code for some of the testing I'm doing.

Also is it normal that the arm library is 7mb, when the windows ones are less then 1? That just seems huge, especially when the HAL is less then 1 MB.

333fred:
Should be fixed, please try again Thad.

build.gradle
```diff
@@ -1,284 +1,149 @@
...
-        if (vs14Dir.exists()) {
-          def vs12Dir = file('C:\\Program Files (x86)\\Microsoft Visual Studio 12.0')
-          if (!vs12Dir.exists()) {
-            throw new GradleException('Compilation with VS 2015 is not currently supported by Gradle. Please install VS 2013, or use the CMake build option.')
-          }
-          installDir = vs12Dir
+ext.getPlatformPath = { binary ->
+    if (binary.targetPlatform.architecture.arm) {
+        return 'Linux/arm'
+    } else if (binary.targetPlatform.operatingSystem.name == 'Linux') {
```
PeterJohnson: On my system, this needs to be if (binary.targetPlatform.operatingSystem.linux) and similarly I think the Windows line needs to be if (binary.targetPlatform.operatingSystem.windows).

333fred:
As to the binary size, I did a quick LOC (lines of code, if you don't know) count for the HAL and ntcore. ntcore is 18244 LOC, and the HAL is 16119 LOC. Of the HAL's code, 9013 LOC are from the NI headers, so we actually have 7106 LOC. So that's why it's so much bigger than the HAL. As to why it's only 1MB on Windows, not sure. Maybe something to do with exporting less symbols for the Windows binary?

ThadHouse:
With those changes it builds properly in Ubuntu 14.04 on my system. It also build properly on Windows, however the SavePersistent/0 and SavePersistent/1 tests fail.

PeterJohnson:
Regarding the arm library size, the main reason is the debugging information is included in the built shared object.  Removing -g3 from the options drops the size from 6M to 2M, and building with -O2 instead of -O0 further reduces the size to just 600K.  We really need a debug and a release build.  Note it's possible on Linux to separate the debug info out, but this is more gradle work.

PeterJohnson:
Hmm, I thought I fixed SavePersistent, but maybe it's MSVC version dependent (see StorageTest.cpp lines 460-464.

build.gradle
333fred: Fixed.

333fred:
This is the output I'm getting from the test:

``` bash
[ RUN      ] StorageTestsPersistent/StorageTestPersistent.SavePersistent/0
C:\Users\333fr\Documents\first\ntcore\test\unit\StorageTest.cpp(461): error: Value of: line
  Actual: { 'd' (100, 0x64), 'o' (111, 0x6F), 'u' (117, 0x75), 'b' (98, 0x62), 'l' (108, 0x6C), 'e' (101, 0x65), ' ' (32, 0x20), '"' (34, 0x22), 'd' (100, 0x64), 'o' (111, 0x6F), 'u' (117, 0x75), 'b' (98, 0x62), 'l' (108, 0x6C), 'e' (101, 0x65), '/' (47, 0x2F), 'b' (98, 0x62), 'i' (105, 0x69), 'g' (103, 0x67), '"' (34, 0x22), '=' (61, 0x3D), '1' (49, 0x31), '.' (46, 0x2E), '3' (51, 0x33), 'e' (101, 0x65), '+' (43, 0x2B), '0' (48, 0x30), '8' (56, 0x38)                   }
Expected: "double \"double/big\"=1.3e+008"
Which is: 00007FF63809BF50
[  FAILED  ] StorageTestsPersistent/StorageTestPersistent.SavePersistent/0, where GetParam() = false (1 ms)
[ RUN      ] StorageTestsPersistent/StorageTestPersistent.SavePersistent/1
C:\Users\333fr\Documents\first\ntcore\test\unit\StorageTest.cpp(461): error: Value of: line
  Actual: { 'd' (100, 0x64), 'o' (111, 0x6F), 'u' (117, 0x75), 'b' (98, 0x62), 'l' (108, 0x6C), 'e' (101, 0x65), ' ' (32, 0x20), '"' (34, 0x22), 'd' (100, 0x64), 'o' (111, 0x6F), 'u' (117, 0x75), 'b' (98, 0x62), 'l' (108, 0x6C), 'e' (101, 0x65), '/' (47, 0x2F), 'b' (98, 0x62), 'i' (105, 0x69), 'g' (103, 0x67), '"' (34, 0x22), '=' (61, 0x3D), '1' (49, 0x31), '.' (46, 0x2E), '3' (51, 0x33), 'e' (101, 0x65), '+' (43, 0x2B), '0' (48, 0x30), '8' (56, 0x38) }
Expected: "double \"double/big\"=1.3e+008"
Which is: 00007FF63809BF50
[  FAILED  ] StorageTestsPersistent/StorageTestPersistent.SavePersistent/1, where GetParam() = true (0 ms)
```

333fred:
Commenting out that ifdef fixes the test for me. I'm running VS2015.

PeterJohnson:
Yep-that means that on MSVC 2015 you're getting the same output as I do on Linux.  It looks like MSVC 2013 and MSVC 2015 behavior is different.  Now to look up what _MSC_VER is on 2013 so I can properly conditionalize the test :)

ThadHouse:
That's the same output I get. Running VS2015 as well.

Glad there's a way to get the linux and arm shared libraries to less then 1MB. With them being 6+MB, my DLL with all 5 build sources included was going to be over 20MB. It's less then 5 MB now with optimizations on.

PeterJohnson:
As the arm library size also directly impacts Java program startup time (as it has to unzip the library to a temp location) we should be defaulting to optimized libs, at least for the .jar, and arguably for C++ use as well.

333fred:
One thing that I am wondering about, which is probably a slightly broader discussion, is whether to ship  builds for teams with the debug info in there. I have to say, I like the idea that teams will be able to step through their code and see what is actually happening under the hood, and not have it just disappear into precompiled binary.

PeterJohnson:
So maybe optimize it for Java and leave debug info in there for C++?  Eesh, I hate to do that as it will make builds ugly.  This is one of the reasons I wanted to make it a .so that shipped with the image.

333fred:
I think we can all agree that is not a workable solution. I like the idea of being able to step through code, but not to the point of causing us that level of pain.

ThadHouse:
I would be happy as well with the .so being shipped with the image. I would need the release builds for all the libraries, because a 20MB assembly is just too large to ship and send over FTP every code build. Java would have the same issue.

Being able to step through the code would be good, but would the Java debugger be smart enough to step into native code?

333fred:
The Java debugger probably wouldn't, I've never tried to debug JNI before. C++ would be the one that benefits, although now that I think about it, if Java isn't going to benefit from this, it's not too much of an issue to remove it from C++ as well.

PeterJohnson:
Building with -O2 will also result in substantially better performance.  Us developers however occasionally benefit from debug info. :)  Fred, are you familiar enough with objcopy et al to know how to do the split-debuginfo thing?  If I give you a command line could you integrate it into gradle?  Basically it takes the .so generated by the linker and splits it into a debuginfo file and a debug-less .so file.

333fred:
I should be able to make it work if you give me a command line for it.

PeterJohnson:
The other thing about placing the .so outside the .jar is that java crashes actually generate useful core files.  Not that it's particularly a concern for teams, but I definitely split it out myself when doing any kind of debugging.

It actually takes 3 commands to split out debug info properly:
objcopy --only-keep-debug libntcore.so libntcore.so.debug
strip -g libntcore.so
objcopy --add-gnu-debuglink=libntcore.so.debug libntcore.so

PeterJohnson:
Just committed a fix for the MSVC 2015 test failure as well.

333fred:
So do we want to have a separate build for debug and release, or just do the objcopy method, or some combination thereof? I have the separate debug and release builds working in a separate branch, just testing on Linux before I update this request.

PeterJohnson:
I think a selectable build with 2 options is sufficient, with the two options being pure-debug (-O0 -g, what we have now) and release-with-separate-debuginfo (-O2 + objcopy).

PeterJohnson:
With the default being release, as that's what we'll want teams to use.

333fred:
Two things:
1. I'm assuming that I need to have the release build still be building with -g, and then do the objcopy+strip?
2. Does Windows also have this issue? I believe Windows already separates its debug info into a separate pdb, so I don't think so?

PeterJohnson:
1. Correct.  You build with -O2 -g and then do the objcopy+strip.
2. No, you're correct that Windows already separates the debug info.  Although release and debug builds might be a good idea there as well (not sure of the right command line options though, as I've always used the IDE to do that).

ThadHouse:
As it is right now, windows builds release builds without debug info when built through gradle.

PeterJohnson:
Thad, do you know the right compiler option incantations to turn on/off debug info / optimizations for Windows?

ThadHouse:
I do not, but I think it's just /DEBUG. It looks like the windows.gradle file already has a setting for building debug. Let me test it out.

333fred:
I already have those set up.

333fred:
It's /DEBUG to the linker, and /Zi /FS to the compiler.

333fred:
Alright, so I've updated the build with the following changes:
1. There is now a debug and release version. Building the jar or ntcoreZip targets will only build the release version.
2. The release version is built with -O2 -g, and the resulting shared object is stripped with the commands shown by Peter above. I've tested this on Arch, and it appears to be working correctly. The main .so is down to 600 KB on my system, with a 6.3 MB debug library next to it.
3. I've tried to do this as simply as possible. On the Linux, Arm, and mac platforms, I generate the objcopy and strip commands programmatically, so hopefully that won't be too fragile. The one thing that I'm concerned with is that due to the Gradle lifecycle, I can't actually access the binary's tasks during the setup, so I have to set up an event handler using an internal API. I don't think there's a better way of doing this, as we only want to hook up the objcopy/strip tasks to this specific binary.

If there's any feedback on this, I'll see it in the morning (at least, more actual morning than it is now. 1 AM doesn't count as morning yet :)

333fred:
Also, as a note, I do not have a Mac, so all of this could (and quite possibly will) break horribly on them.

PeterJohnson:
Great!  Testing now.  On a separate topic, is this a warning to be concerned about? "Defining custom 'check' task when using the standard Gradle lifecycle plugins has been deprecated and is scheduled to be removed in Gradle 3.0"

PeterJohnson:
Something's not quite right; if you just do "gradlew build" the resulting ntcore-linux.jar has both the release and debug .so's embedded:
        0  2015-10-05 22:25   Linux/
        0  2015-10-05 22:25   Linux/amd64/
  6157032  2015-10-05 22:21   Linux/amd64/libntcore.so
   664776  2015-10-05 22:25   Linux/amd64/libntcore.so
        0  2015-10-05 22:25   Linux/x86/
  5773012  2015-10-05 22:22   Linux/x86/libntcore.so
   649536  2015-10-05 22:25   Linux/x86/libntcore.so

Well, that and it builds both the release and debug versions and runs the tests on both, which means it runs the test suite 4 times.  Can you make the "build" command just build the release versions and make the debug ones require a property to be set similar to how buildPlatform works?

PeterJohnson:
To clarify, the issue with "gradlew build" is if you run it twice.  The first run puts only the release files in the jar, but the second run adds the debug .so's to the jar.

ThadHouse:
Windows is doing the same thing as well, when calling gradlew build. gradew jar however does not have the issue.

333fred:
Unfortunately, there is no way (that I have found, at least) to modify the build task. The build task is designed to build absolutely everything. To remedy this, I've created two other build tasks, called releaseBuild and debugBuild. These do exactly what it sounds like: they build all the release or all the debug binaries. I've also created two new packaging tasks: debugNtcoreZip and debugJar. The nctoreZip tasks (debug and not debug) create a zip of all the created binaries, including both shared objects and static libraries. The jar thing was a bug, that's fixed, and I've also added a second jar task, debugJar, which creates a jar with the debug versions of the binaries.

333fred:
Also, the debugBuild and releaseBuild will run the x86 and x64 tests for their respective build types, so each will only do 2 test runs. The jar tasks and zip tasks will do the same.

333fred:
The check thing is because I'm defining a task check() at the top of the test.gradle class, resolving an exception that gets thrown otherwise. https://discuss.gradle.org/t/plugin-google-test-causes-failure-when-plugin-java-is-applied/10895

PeterJohnson:
This just feels a little backwards to me right now.  We seldom want to build both release and debug, but you can do that with "gradlew build"; we often want to build both arm and native, but that requires running both "gradlew -PbuildPlatform=arm build" and "gradlew build".  How I think it should work is exactly the opposite: "gradlew build" builds arm and native release, and "gradlew -PbuildType=debug build" builds arm and native debug.  If the arm build thing is hard to do, maybe we don't do that, but is implementing "debugBuild" versus "-PbuildType=debug build" that much different?

333fred:
I mean, that's not too hard, but I'm still having trouble thinking of a scenario when we want to build both arm and x86/x64. If you want arm, you're probably going to be building as part of WPILib proper. If you're not building as part of WPILib, then you're probably building for the current platform. Do you see another scenario that I'm missing?

PeterJohnson:
Good point since it's generally embedded in other projects... and I guess the wpilib simulation case will probably be separate build commands anyway?  Maybe I'm the only one who really builds ntcore all by itself :)... and I can easily run two gradle commands when the need arises.

I would like to see the debug/release thing done as a -PbuildType option though, because otherwise we will also pollute e.g. the wpilib lib with building both debug and release, when only the release one is actually desired.

One other thing: wpilib is built as a static library too... do we need to do anything in ntcore to support this?

ThadHouse:
I'll be building ntcore by itself too, but I'll need Win 32 and 64, Linux 32 and 64, and Arm, so 3 builds are barely any worse then 2. Plus hopefully there will be minimal updates during the actual build season to the ntcore library. I (and I'm assuming Python) would only need the zips to be built as well, which are much quicker then building the whole build.

The gradle builds already build the static library for all 5 system types I've tried, so there shouldn't be anything that needs to be done.

ThadHouse:
One more thing I just remembered. If I tried just running gradlew zip, or any of the individual library builds, after a fresh clone, the build fails since it can't find the JNI headers. Can the JNI header building be added as a dependency of any build?

333fred:
Wait, what? The jni headers are already integrated into the build. What command line are you using?

ThadHouse:
```
:installX86ReleaseNtcoreTestGoogleTestExe
:linkX64ReleaseNtcoreSharedLibrary
ntcore.def : error LNK2001: unresolved external symbol JNI_OnLoad
ntcore.def : error LNK2001: unresolved external symbol JNI_OnUnload
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_addConnectionListener
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_addEntryListener
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_callRpc
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_containsKey
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_deleteAllEntries
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_deleteEntry
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_flush
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_forcePutBoolean
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_forcePutBooleanArray
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_forcePutDouble
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_forcePutDoubleArray
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_forcePutRaw
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_forcePutString
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_forcePutStringArray
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getBooleanArray__Ljava_lang_String_2
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getBooleanArray__Ljava_lang_String_2_3Z
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getBoolean__Ljava_lang_String_2
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getBoolean__Ljava_lang_String_2Z
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getConnections
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getDoubleArray__Ljava_lang_String_2
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getDoubleArray__Ljava_lang_String_2_3D
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getDouble__Ljava_lang_String_2
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getDouble__Ljava_lang_String_2D
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getEntries
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getEntryFlags
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getRaw__Ljava_lang_String_2
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getRaw__Ljava_lang_String_2_3B
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getRpc__Ljava_lang_String_2
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getRpc__Ljava_lang_String_2_3B
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getStringArray__Ljava_lang_String_2
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getStringArray__Ljava_lang_String_2_3Ljava_lang_String_2
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getString__Ljava_lang_String_2
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getString__Ljava_lang_String_2Ljava_lang_String_2
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getType
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getValue__Ljava_lang_String_2
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getValue__Ljava_lang_String_2Ljava_lang_Object_2
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_loadPersistent
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_now
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_putBoolean
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_putBooleanArray
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_putDouble
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_putDoubleArray
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_putRaw
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_putString
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_putStringArray
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_removeConnectionListener
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_removeEntryListener
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_savePersistent
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_setEntryFlags
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_setLogger
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_setNetworkIdentity
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_setUpdateRate
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_startClient
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_startServer
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_stopClient
ntcore.def : error LNK2001: unresolved external symbol Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_stopServer
C:\Users\Thad\Documents\GitHub\newntcore\ntcore\build\binaries\ntcoreSharedLibrary\x64Release\ntcore.lib : fatal error LNK1120: 58 unresolved externals

:linkX64ReleaseNtcoreSharedLibrary FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':linkX64ReleaseNtcoreSharedLibrary'.
> A build operation failed.
      Linker failed while linking ntcore.dll.
  See the complete log at: file:///C:/Users/Thad/Documents/GitHub/newntcore/ntcore/build/tmp/linkX64ReleaseNtcoreSharedLibrary/output.txt

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
```

I get that when attempting to run gradlew ntcoreZip, or any of the x64(86)Release(Debug)NtcoreSharedLibrary functions. 

Actually I think this is failing because I have been running -PskipJava since my laptop doesn't have the JDK. But I think we should be able build the shared libraries without having the JDK installed.

This also is a windows only issue, since the linux libraries don't explicitly link to a .def file.

PeterJohnson:
Er, well, if you don't have a native JDK, the native shared object _can't_ get built with the JNI interfaces (we include the arm JNI headers in the source tree but not the native ones).  To support this scenario, we'll need to make a custom ntcore.def file that leaves off the JNI symbols and is only used in the skipJava case.

Right now, the tree also does not allow cross-build of an arm JNI shared object if you don't have a local JDK, because it needs to run Java to create the appropriate .h files.  This we could fix by simply adding the generated .h files to the tree.

ThadHouse:
It's probably not worth it to do 2 separate def files. Its not that hard to install the JDK especially on windows. Linux will properly build and run even without the JNI symbols as long as you don't need them, and anyone compiling this themselves for non java reasons probably won't need the Java exports.

333fred:
Another small update code update, but with larger potential behind it. First, the update: a few general cleanup details. The debug build has been moved back to a build property. To produce a debug build, use -Pdebug. I've changed the buildPlatform setting to simply be armBuild. buildPlatform implied that you could do -PbuildPlatform=linux from Windows and have it work. armBuild is a simple "does it exist" check, so just -ParmBuild from the command line will suffice. I've tested this on Windows, but I'm currently in Roatan, Honduras, and away from my Linux machine.

Brad and I have had a lot of free time here to talk about what we want the future of wpilib builds to look like (we apparently have nothing better to do on dive trips). We've decided that we're going to remove the submodule from the wpilib repo and instead have dependency resolution done via Maven. How exactly this will work is something I'm currently designing, but the main thing we want to get away from is having to maintain the submodule. There's already been an accidental commit that reverted the submodule commit to a previous version, causing compilation errors, and I'm getting constant questions about why a fresh clone doesn't seem to work. We need to get the Maven stuff up and running anyway, and the work that's done here should make it smoother for everyone.

333fred:
The publishing steps are now added. We can publish to 4 different maven repositories, depending on variables defined on the command line.

PeterJohnson:
The latest patchset doesn't seem to be pulling out the debuginfo on the release builds?  After running just "gradlew build", the binaries directory contains an unstripped .so, and so does the jar.  It looks like the previous way you hooked objcopy in doesn't work for some reason?

Also, on the patchset, could you make sure the files have EOL at EOF?  A few of the files don't.

333fred:
Both should be resolved.

PeterJohnson:
Looks good to me in terms of Linux builds.  Have you tested Windows?  Is there any reason for me not to merge this?

333fred:
I've done all my development and testing on Windows, which is probably why I didn't notice the missing objcopy step. Should be good to merge, and in fact Brad and I were hoping to get this merged in today. Don't worry about updating the allwpilib submodule reference, as that reference will be going away.

PR #33, opened 2015-11-18
All Double usage now supports the abstract Number type

JLLeitschuh:


PeterJohnson:
For putNumber and getNumber I would prefer to add an overload rather than replacing the double version entirely.  Otherwise putting a double requires two conversions.  Or is the issue that doing that would cause Java to see that as an ambiguous overload when a raw double is used?

JLLeitschuh:
Since Double extends Number it would be ambiguous which one should be called in that case correct?

JLLeitschuh:
Are you talking about the parameter or the return type?

PeterJohnson:
Parameter.  And I'm talking about double not Double.

333fred:
Maybe a little bit late to comment on the issue, but now that school's done I'm looking around at things I haven't been paying attention to. I'm not a fan of replacing Double with Number in any fashion. The NT spec says that all numbers are IEEE double-precision floating point numbers, and the parameter types here make that clear. Changing it to number makes it unclear what type is being used under the hood, when in fact the type that's being used is very explicitly defined.

PR #37, opened 2015-12-19
Fixes ContainsKey

nekoknot:
Since C++11, `shared_ptr` has an explicit `operator bool` and thus can't be implicitly converted. The result is a `ContainsKey` which always returns `false`.

This fixes it by explicitly checking against `nullptr`.


PeterJohnson:
Thanks!  Committed a slightly different fix and a testcase.

nekoknot:
Just my opinion, but I think a comparing it to `nullptr` or even using the `.bool()` function would be much clearer than what it's doing now. Fundamentally it's a null check, not a true/false check. Even though the end result is the same, code should convey intent.

PR #44, opened 2016-01-04
Fixes Android Build issues

ThadHouse:
Trying to build with the android standalone compiler, and these 2 things were causing errors. I don't know what an equivalent to basename would be, because I don't really know what it does.

### DIFF
--- a/src/Log.cpp
+++ b/src/Log.cpp
@@ -18,6 +18,10 @@
  #include <libgen.h>
 #endif
 
+#ifdef __ANDROID__
+ #include <libgen.h>
+#endif
+
 using namespace nt;
 
 ATOMIC_STATIC_INIT(Logger)
--- a/src/networktables/NetworkTable.cpp
+++ b/src/networktables/NetworkTable.cpp
@@ -44,7 +44,8 @@ void NetworkTable::SetTeam(int team) {
 #ifdef _MSC_VER
   sprintf_s(tmp, "roboRIO-%d-FRC.local\n", team);
 #else
-  std::snprintf(tmp, 30, "roboRIO-%d-FRC.local\n",team);
+  using namespace std;
+  snprintf(tmp, 30, "roboRIO-%d-FRC.local\n",team);
 #endif
   SetIPAddress(tmp);
 }


PeterJohnson:
basename converts a filename with path (e.g. /foo/bar.txt) into just the filename (bar.txt).  The reason it's used here is to make error messages less verbose.

PeterJohnson:
Please change this to use `#elif defined(__ANDROID__)` instead of just `#elif __ANDROID__`.  The latter has issues on some compilers when `__ANDROID__` is not defined.

ThadHouse:
Ok changed to use defined.

333fred:
From what I'm seeing in the ndk includes, libgen.h should exist, along with both `basename` and `basename_r`. They should exist in platforms/android-12/choose an architecture/usr/include

333fred:
And actually, every android version from 12 and up.

ThadHouse:
Ah. Added an ifdef to define libgen.h if it sees android.

src/Log.cpp
```diff
@@ -51,6 +55,9 @@ static void def_log_func(unsigned int level, const char* file,
   std::fprintf(stderr, "NT: %s: %s (%s:%d)\n", levelmsg, msg, basename(basestr),
                line);
   delete[] basestr;
+#elif  defined(__ANDROID__)
+  std::fprintf(stderr, "NT: %s: %s (%s:%d)\n", levelmsg, msg, basename(file),
+               line);
```
PeterJohnson: Can't this section be deleted now since it's the same as the #else section?

src/networktables/NetworkTable.cpp
```diff
@@ -1,7 +1,6 @@
 #include "networktables/NetworkTable.h"
 
 #include <algorithm>
-
```
PeterJohnson, line 4: Removing this line is unnecessary and doesn't match style of other files.

src/networktables/NetworkTable.cpp
```diff
@@ -43,6 +42,8 @@ void NetworkTable::SetTeam(int team) {
   char tmp[30];
 #ifdef _MSC_VER
   sprintf_s(tmp, "roboRIO-%d-FRC.local\n", team);
+#elif defined(__ANDROID__)
+  snprintf(tmp, 30, "roboRIO-%d-FRC.local\n",team);
```
PeterJohnson: std::snprintf is supposed to be in C++11.  Maybe try adding `#include <cstring>`?

src/Log.cpp
ThadHouse: True. I'll fix this tonight.

src/networktables/NetworkTable.cpp
ThadHouse, line 4: Whoops didn't realize that got removed. I'll fix it.

src/networktables/NetworkTable.cpp
ThadHouse: I'll try that. I tried including `cstdio` and `stdio` without those working, but I'll try `cstring` when I get home. Is it save to include `cstring` for everyone, or does that have to be in its own `#ifdef`?

PeterJohnson: It's safe to include for everyone.  If that doesn't work, we could add `using namespace std;` to the `#else` section and remove the `std::` to get common code across the different platforms.

333fred: It does look like the cstdio header is included with the various C++ implementations that come with the NDK. Android's default C++ support is _very_ limited, and I'm actually kind of surprised everything else compiles. Thad, are you using one of the helper runtimes? If so, which one? It would be nice to see some documentation updates so others don't have to do the same learning process you've been having to.

ThadHouse: I used the directions here http://developer.android.com/ndk/guides/standalone_toolchain.html to create a standalone toolchain for android. I believe I used this switch `--toolchain=arm-linux-androideabi-4.9` to get the 4.9 version I was using. Then what I did was placed the standalone toolchain in my path, and then changed the arm.gradle file to use the different toolchain prefix. Then just compiled with gradle, and it worked (well kind of). I tested on my nexus 6 with Xamarin, and it worked and connected. However, MDNS did not work, and for some reason the code had be in debug mode, as release mode would not connect to the network properly. 

I was kind of surprised that it worked too. But the changes here are the only changes I needed to get it working and compiling.

ThadHouse: It would be nice if we could have another option in gradle that would let android be built without a gradle change. It would probably require a gradle option to specify the toolchain prefix, as different versions have different prefixes. However I don't know enough about gradle to set this up if we did end up wanting to do this.

333fred: I don't know what you mean by "be build without a gradle change". What is being built? What is a "gradle change"?

ThadHouse: To get ntcore to build using the standalone android compiler, I had to use the arm settings, but change the prefix for the toolchain from `arm-frc-linux-gnueabi-` to `arm-linux-androideabi-`. It would be nice to have an android.gradle and an android platform in the gradle configuration. That way we could just specify `-Pandroid`, and perform an android build.

333fred:
Taking this conversation out of line notes.

I don't know as though that's something we should be putting in the build files. By putting it there, we imply that we actively support compiling for Android, when we don't, and as far as I'm aware have no plans to do so. @PeterJohnson, thoughts on that?

PeterJohnson:
I agree that we shouldn't add android-specific gradle files.  However, I'm thinking a reasonable compromise would be to allow compilerPrefix in arm.gradle to be changed with a `-P` option... that way the arm.gradle file could be used with any ARM target compiler variant.

333fred:
Alright. Opened #45, I'll probably get to that sometime tonight.

ThadHouse:
I fixed all the issues in the line comments, and tested to make sure it compiled both with the android compiler and the roboRIO compiler.

PR #48, opened 2016-01-19
Load ntcore from path if extracted load fails

ThadHouse:
There's a lot of buzz going around the internet about people trying to
get ntcore working on other devices. One of the things that makes it
harder is having to have a separate jar for each platform. What this
change does is if the loading of the extracted library fails, it will
attempt to load ntcore from the path. This means that a program like
GRIP can just provide the libntcore.so and not have to worry about
compiling different jars for different platforms, and instead just provide different platforms. There might be a better workaround for future seasons, but for now this should help make things easier.

### DIFF
--- a/java/src/edu/wpi/first/wpilibj/networktables/NetworkTablesJNI.java
+++ b/java/src/edu/wpi/first/wpilibj/networktables/NetworkTablesJNI.java
@@ -51,8 +51,11 @@ public class NetworkTablesJNI {
             os.close();
             is.close();
           }
-
-          System.load(jniLibrary.getAbsolutePath());
+          try {
+            System.load(jniLibrary.getAbsolutePath());
+          } catch (UnsatisfiedLinkError e) {
+            System.loadLibrary("ntcore");
+          }
         } else {
           System.loadLibrary("ntcore");
         }


PeterJohnson:
It does already fall back if it can't find the library in the .jar (see line 57 in the original file).  The only time this can fail is if there's a .so in the correct location (e.g. /Linux/arm) but it can't be loaded for some other reason (e.g. linkage error or incorrect sub-architecture).

ThadHouse:
Maybe I named the PR wrong. The main reason I wanted to add it was so it would be able to load a different native library easier if the built in one failed to load properly. Mainly for things like Raspberry Pis, Beaglebone Blacks, and NVidia Jetsons, which are not compatible with the RoboRIO binary. For now it might be easier to just enable us to load any native library, rather then having to rebuild and redistribute the entire jar for another architecture. 

Later it might be good to integrate other native libraries into the main jar and check for the RoboRIO vs other platforms, but it would be nice to have a workaround other then compiling a completely separate jar, which currently would make GRIP much harder to use on other devices, at least the way its currently written.

PeterJohnson:
I'm basically saying that you can just delete the .so's out of the existing .jar and it will do exactly that today (fall back to trying to load libntcore.so from path).  This patch just makes it a little more robust.  I'm not saying I won't merge it, I'm just saying you can achieve the same goals without it :)

ThadHouse:
True. I didn't think about that. 

I do wish I could get a cross compiler working that didn't require a custom libstdc++.so.6. That's whats stalling a ArmV7 HF build right now. The only way I can get rid of that dependency is to natively compile, which isn't really conducive to build server use. Do you know  of any other cross compilers, or a way to specify cxxabi_1.3.8 vs cxxabi_1.3.9 in gcc?

PeterJohnson:
You need to use a different version of gcc to use a different cxxabi.  For example, I can cross-compile to the roboRio's native libstdc++.so.6.0.18 using g++ 4.8.2.  It looks like cxxabi_1.3.8 is gcc 4.9.0, while 1.3.9 is gcc 5.1.0 (see https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html).

PR #51, opened 2016-02-06
Adds extended Remote Connection Listener

ThadHouse:
Since we now get ConnectionInfo when setting a connection listener,
there is some good information in there that teams could use.
Implemented using default interface methods, so teams should see no
change if they don't implement the Ex methods.

I noticed that the connection listener methods don't exist at all in
C++, so they did not get added there.

Sorry about the 2 extra commits. My master was based off of gerrit, which has those 2 commits.


PeterJohnson:
I pushed the gerrit changes up to master here so you should be good to rebase.  Also, it looks like there's a mix of tabs and spaces in your IRemoteConnectionListener.java changes?  These should be spaces only.

ThadHouse:
Ok I got all that stuff fixed. Also fixed the Javadocs for the new methods.

Issue #53, opened 2016-02-15
Naming for new atomic get set operation


ThadHouse:
In the past, SmartDashboard has always had the issue where you had to put a number to the dashboard before you could read it and change values from the SmartDashboard. Most teams I know would just run a bunch of Put**\* calls in RobotInit, that would write safe defaults, and then could be safely read from robot code and edited from the dashboard. However, the addition of persistent entries has made this a much harder task. The Put calls in RobotInit are no longer safe, as they very likely will override the value read from persistent. We've had this happen to us, where we set values persistent, run testing for an hour, and then when we reboot all the persistent values are overridden because we forgot to remove the put calls in RobotInit. I thought about first reading to check if the key exists, and only putting if it does not, however that's not thread safe, and the SmartDashboard class in WPILib does not have a built in method to check if key exists (I plan on pushing some patches to WPILib later to add some of the new NT3 functionality).

I have been working on an atomic operation to first check if the key exists, returning the value in the table if it does, otherwise setting the table to a passed in value, and then returning that. I think I have the code working, and wrote working tests, however I don't really know what a good name for a method like that would be. Currently I have it set to `GetSetEntryValue`, however that doesn't sound right, and I can't come up with anything better. Any ideas? Also do you think this would be a valid addition? You can see the code here https://github.com/ThadHouse/ntcore/commit/9f2de773cddf151bbab5f05692806ef9ab2159bc

PeterJohnson:
In Python dictionaries, this type of operation is called SetDefault.   "setdefault(key, default): If key is in the dictionary, return its value. If not, insert key with a value of default and return default."  Yes, I think this is a valid addition.

PR #54, opened 2016-02-16
Adds SetDefault methods to NetworkTables

ThadHouse:
There was no way to atomically check for a key in the table,
and then setting if it if non existant. Back before persistent
this was not a problem, however now it is, as its possible for
values to be added before team's robot programs start. This makes
the old method of calling Put**\* methods in RobotInit invalid.
This adds SetDefault methods, which do this atomically.

Discussed in #53

### DIFF
--- a/include/networktables/NetworkTable.h
+++ b/include/networktables/NetworkTable.h
@@ -267,6 +267,14 @@ class NetworkTable : public ITable {
    * @return False if the table key already exists with a different type
    */
   bool PutNumber(llvm::StringRef key, double value) override;
+  
+  /** 
+   * Gets the current value in the table, setting it if it does not exist.
+   * @param key the key
+   * @param defaultValue the default value to set if key doesn't exist.
+   * @returns False if the table key exists with a different type
+   */
+  virtual bool SetDefaultNumber(llvm::StringRef key, double defaultValue) override;
 
   /**
    * Gets the number associated with the given name.
@@ -301,6 +309,15 @@ class NetworkTable : public ITable {
    * @return False if the table key already exists with a different type
    */
   virtual bool PutString(llvm::StringRef key, llvm::StringRef value) override;
+  
+  /** 
+   * Gets the current value in the table, setting it if it does not exist.
+   * @param key the key
+   * @param defaultValue the default value to set if key doesn't exist.
+   * @returns False if the table key exists with a different type
+   */
+  virtual bool SetDefaultString(llvm::StringRef key, 
+                                llvm::StringRef defaultValue) override;
 
   /**
    * Gets the string associated with the given name.
@@ -336,6 +353,14 @@ class NetworkTable : public ITable {
    * @return False if the table key already exists with a different type
    */
   virtual bool PutBoolean(llvm::StringRef key, bool value) override;
+  
+  /** 
+   * Gets the current value in the table, setting it if it does not exist.
+   * @param key the key
+   * @param defaultValue the default value to set if key doesn't exist.
+   * @returns False if the table key exists with a different type
+   */
+  virtual bool SetDefaultBoolean(llvm::StringRef key, bool defaultValue) override;
 
   /**
    * Gets the boolean associated with the given name.
@@ -376,6 +401,15 @@ class NetworkTable : public ITable {
   virtual bool PutBooleanArray(llvm::StringRef key,
                                llvm::ArrayRef<int> value) override;
 
+  /** 
+   * Gets the current value in the table, setting it if it does not exist.
+   * @param key the key
+   * @param defaultValue the default value to set if key doesn't exist.
+   * @returns False if the table key exists with a different type
+   */                             
+  virtual bool SetDefaultBooleanArray(llvm::StringRef key,
+                                      llvm::ArrayRef<int> defaultValue) override;
+
   /**
    * Returns the boolean array the key maps to. If the key does not exist or is
    * of different type, it will return the default value.
@@ -403,6 +437,15 @@ class NetworkTable : public ITable {
   virtual bool PutNumberArray(llvm::StringRef key,
                               llvm::ArrayRef<double> value) override;
 
+  /** 
+   * Gets the current value in the table, setting it if it does not exist.
+   * @param key the key
+   * @param defaultValue the default value to set if key doesn't exist.
+   * @returns False if the table key exists with a different type
+   */
+  virtual bool SetDefaultNumberArray(llvm::StringRef key,
+                                     llvm::ArrayRef<double> defaultValue) override;
+
   /**
    * Returns the number array the key maps to. If the key does not exist or is
    * of different type, it will return the default value.
@@ -426,6 +469,15 @@ class NetworkTable : public ITable {
   virtual bool PutStringArray(llvm::StringRef key,
                               llvm::ArrayRef<std::string> value) override;
 
+  /** 
+   * Gets the current value in the table, setting it if it does not exist.
+   * @param key the key
+   * @param defaultValue the default value to set if key doesn't exist.
+   * @returns False if the table key exists with a different type
+   */
+  virtual bool SetDefaultStringArray(llvm::StringRef key, 
+                                     llvm::ArrayRef<std::string> defaultValue) override;
+
   /**
    * Returns the string array the key maps to. If the key does not exist or is
    * of different type, it will return the default value.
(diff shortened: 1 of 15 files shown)


include/networktables/NetworkTable.h
```diff
@@ -295,6 +303,15 @@ class NetworkTable : public ITable {
...
   virtual bool PutString(llvm::StringRef key, llvm::StringRef value) override;
+  
+  /** 
+   * Gets the current value in the table, setting it if it does not exist.
+   * @param key the key
+   * @param defaultValue the default value to set if key doens't exist.
+   * @returns the key in the table, or the defaultValue if it doesn't exist.
+   */
+  std::string SetDefaultString(llvm::StringRef key, 
+                                   llvm::StringRef defaultValue);
```
333fred: Please standardize your next-line format. Everything should be at the same indent as the previous args, like the existing functions in the file.

include/networktables/NetworkTable.h
```diff
@@ -295,6 +303,15 @@ class NetworkTable : public ITable {
    * @return False if the table key already exists with a different type
    */
   virtual bool PutString(llvm::StringRef key, llvm::StringRef value) override;
+  
+  /** 
+   * Gets the current value in the table, setting it if it does not exist.
+   * @param key the key
+   * @param defaultValue the default value to set if key doens't exist.
```
333fred: Typo.

include/ntcore_c.h
```diff
@@ -789,6 +807,40 @@ NT_String *NT_GetEntryStringArray(const char *name, size_t name_len,
                                   unsigned long long *last_change,
                                   size_t *arr_size);
 
+/* Set Default Values */
+
+
+int NT_SetDefaultEntryBoolean(const char *name, size_t name_len,
```
333fred, line 830: These should all have documentation.

include/networktables/NetworkTable.h
ThadHouse: So on some I can do that, but on some the indents like that go over 80 characters wide. Whats the best solution for methods with arguments that physically cant fit in 80 characters? For instance SetDefaultStringArray, and most of the auto val = nt::SetDefaultEntryValue functions.

333fred: I'd say whatever clang-format does. See this gerrit change from Tyler: https://usfirst.collab.net/gerrit/#/c/1222/.

333fred:
A couple of comments:
First, in the case that the key already exists and is a different type, this code will return the default value. I'm not sure this is what we want here; there's no way for a programmer to differentiate between failure and success condition here, and it's likely a sign of programmer error. Perhaps returning NULL, such as what the other getters do, is more appropriate.
This is one of those types of commits that I think I'd prefer if we waited until after the season to actually merge, just in case we have to do another release this season. It's too much new functionality that hasn't been beta tested.

ThadHouse:
We could return null from the Storage methods, but then what do we do for the user facing NetworkTables methods? For the array and string ones it's not really a problem, but we can't return a null bool or double. Plus in java, we can't pass a status parameter by reference either, so detecting if that was success is going to be hard without a separate class.

Plus, the main use case I can think of for this is for teams to put it in RobotInit, where technically they won't need the return value. The main reason I can think of is so teams don't accidentally overwrite their persistent values.

333fred:
We should do whatever it is we already do for the Get\* methods.

ThadHouse:
They return the default value if the key type doesn't match. (Or throw exceptions, which I don't think is what we want.)

333fred:
Hmm. I guess the thing is, the semantic meaning of "default value" is different here. Here, we are trying to initialize a value if it does not exist, whereas the Get with a default methods are providing a "give me this if there's an error". The default value being of a different is an error here, but is there in case of an error there.

ThadHouse:
The only way I could see to solve this would be either status parameters, which we can't do in Java, or exceptions, which just seem ugly, and not great on the C++ side. 

The methods were modeled off of python's dictionary SetDefault, which just returns the default value in case of error. I honestly don't know what the right course of action is here. 

What do you think, @PeterJohnson

PeterJohnson:
For consistency with Python, the return type would have to be a variant type.  I don't think it makes sense to return the default unless the default was actually set (which isn't the case if there was a type mismatch, unless we want this to override the type, which we probably don't).  I'm thinking we should just return a boolean like the other Set() methods do.. basically indicating whether or not the value was actually set.  If we really want to, we can make it an int/enum to indicate whether it wasn't set because of a type mismatch or a value already being set, but I think that's overkill. The user can always call Get() to get the current value after calling SetDefault.

ThadHouse:
Ok I got everything changed to return a bool if it was set correctly. Also finished up the docs and tests for this as well.

test/unit/StorageTest.cpp
```diff
@@ -288,6 +288,106 @@ TEST_P(StorageTestEmpty, SetEntryValueEmptyValue) {
...
+    EXPECT_EQ(0xffffu, msg->id());  // not assigned as client
+  EXPECT_EQ(0u, msg->seq_num_uid());
+  EXPECT_EQ(value, msg->value());
+  EXPECT_EQ(0u, msg->flags());
+}
+
+TEST_P(StorageTestPopulateOne, SetDefaultEntryExistsSameType) {
+  // existing entry
+  auto value = Value::MakeBoolean(true);
+  auto ret_val = storage.SetDefaultEntryValue("foo", value);
```
333fred, line 316: How are you making sure that "foo" already exists? gtest does not guarantee test ordering.

ThadHouse, line 316: The StorageTestPopulateOne is actually placing that entry in. I don't know enough about gtest, but its how all the other StorageTests place entries in before testing.

ThadHouse:
I think I got everything fixed in this, so it should be ready to commit. I added the new C methods to the Def file, and also skipped 86 as that one is going to be defined in the round robin commit when that gets merged.

PeterJohnson:
The functions added to NetworkTable should also be added to ITable.

ThadHouse:
Ok. I should be able to do that tonight. Also as a reminder to me SetDefaultString in C++ is not marked virtual like the rest, so remember to change that.

ThadHouse:
Ok all the new functions should be added to ITable. Javadocs were moved to use inheritdoc as well.

src/Storage.cpp
```diff
@@ -474,6 +474,48 @@ std::shared_ptr<Value> Storage::GetEntryValue(StringRef name) const {
...
+  if (m_server && entry->id == 0xffff) {
+    unsigned int id = m_idmap.size();
+    entry->id = id;
+    m_idmap.push_back(entry);
+  }
+
+  // notify (for local listeners)
+  if (m_notifier.local_notifiers()) {
+    // always a new entry if we got this far
+      m_notifier.NotifyEntry(name, value, NT_NOTIFY_NEW | NT_NOTIFY_LOCAL);
```
PeterJohnson: Indentation oops?

src/ntcore_c.cpp
```diff
@@ -692,6 +698,57 @@ NT_String *NT_GetValueStringArray(const struct NT_Value *value,
...
+int NT_SetDefaultEntryBoolean(const char *name, size_t name_len,
+                              int default_boolean) {
+  return nt::SetDefaultEntryValue(StringRef(name, name_len),
+                                  Value::MakeBoolean(default_boolean != 0));
+}
+
+int NT_SetDefaultEntryDouble(const char *name, size_t name_len,
+                             double default_double) {
+  return nt::SetDefaultEntryValue(StringRef(name, name_len),
+                                    Value::MakeDouble(default_double));
```
PeterJohnson: Indentation?

src/ntcore_c.cpp
```diff
@@ -692,6 +698,57 @@ NT_String *NT_GetValueStringArray(const struct NT_Value *value,
...
+                                   const int *default_value, 
+                                   size_t default_size) {
+  return nt::SetDefaultEntryValue(StringRef(name, name_len),
+                                  Value::MakeBooleanArray(llvm::makeArrayRef(default_value, default_size)));
+}
+
+int NT_SetDefaultEntryDoubleArray(const char *name, size_t name_len, const double *default_value,
+                                  size_t default_size) {
+  return nt::SetDefaultEntryValue(StringRef(name, name_len),
+                                    Value::MakeDoubleArray(llvm::makeArrayRef(default_value, default_size)));
```
PeterJohnson: Indentation

ThadHouse:
Fixed indentations. Looked though and couldn't see any more.

PeterJohnson:
One last thing--can you rebase?  I merged the server round robin patch yesterday.

ThadHouse:
Ok. It's rebased.

java/lib/NetworkTablesJNI.cpp
```diff
@@ -917,6 +917,90 @@ JNIEXPORT jobjectArray JNICALL Java_edu_wpi_first_wpilibj_networktables_NetworkT
 
 /*
  * Class:     edu_wpi_first_wpilibj_networktables_NetworkTablesJNI
+ * Method:    setDefaultBoolean
+ * Signature: (Ljava/lang/String;Z)Z
+ */
+JNIEXPORT jboolean JNICALL Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_setDefaultBoolean__Ljava_lang_String_2Z
```
PeterJohnson: I'm confused as to why these have the args suffix when they're not overrides.

ThadHouse: Because I wrote these when I didn't know how JNI worked well and thought I needed them. Fixed.

PR #66, opened 2016-05-22
Added note that multilib GCC is required when building the native version on 64 bit Linux

calcmogul:
Without a GCC installation that can build both 32 and 64 bit executables, the native build will fail on 64 bit Linux machines with an error about unsigned __int128 being undefined. GCC does not support __int128 on 32 bit targets, and GCC was using it via a standard library implementation header intended for 64 bit machines.

---

This change is [Reviewable](https://reviewable.io/reviews/wpilibsuite/ntcore/66)


JLLeitschuh:
Does this need a "in layman's terms" version? IE. what do I need to do to get multilib?

calcmogul:
I added a bit explaining where one can get a multilib-enabled GCC on Linux.

PR #71, opened 2016-06-11
Don't delete persistent entries in DeleteAllEntries.

PeterJohnson:
This is a behavior change but without it DeleteAllEntries is dangerous and
not very useful, so I consider the current behavior to be a bug.

---

This change is [Reviewable](https://reviewable.io/reviews/wpilibsuite/ntcore/71)


ThadHouse:
LGTM. ProcessIncoming looks like it should properly handle the case of a deletion happening and then an update from a remote happening before said remote has received the delete message.

PR #75, opened 2016-06-26
Update Travis OS X build to use newest image.

### LINKED ISSUE #74: Travis OSX Image now has java 8
You can see how I fixed this in GRIP here:

This will speed up build time.

https://github.com/WPIRoboticsProjects/GRIP/pull/608

PeterJohnson:
Fixes #74.


JLLeitschuh:
Looks good to me!

PeterJohnson:
@JLLeitschuh: did you see my comment re: this taking 2x the time to build on Travis (6 mins vs 3 mins)?  If it still makes sense (because we want to move to Java 8 builds) that's fine, but it's not in keeping with your original bug report which seemed to indicate this would be faster.

JLLeitschuh:
Uhhh. I didn't realize this was taking longer. Are you sure this is the case?

PeterJohnson:
https://travis-ci.org/wpilibsuite/ntcore/builds/140316856 (is)
Vs
https://travis-ci.org/wpilibsuite/ntcore/builds/138533053 (was)

OS X build time went from 3 to 6 min.

PR #79, opened 2016-06-30
Move common utility classes to wpiutil library

PeterJohnson:
Note: breaks dependencies (e.g. allwpilib).  Do not merge until this is addressed.


333fred:
Looking pretty good from a Gradle side. A couple of notes:
1. We should find all places that we're using relative paths and replace them with rootDir. I think you're missing the windows def file right now, that's why appveyor failed.
2. build.gradle is starting to get rather large. Do we want to consider factoring it out into multiple files? I'm not sure where the cleanest breaks would be, and I'd love to get feedback from other wpilib developers who've only had minimal interaction with the build files before. Where do you start to get lost when looking at this?

PeterJohnson:
Fixed the builds and split wpiutil and ntcore into separate gradle files.

PR #81, opened 2016-07-03
Notifier: maintain freelist to reuse uids.

### LINKED ISSUE #80: Memory leak on NetworkTable::AddTableListener/NetworkTable::RemoveTableListener
Test code:

``` c++
class BasicListener : public ITableListener {
    void ValueChanged(ITable* source,
        llvm::StringRef key,
        std::shared_ptr<nt::Value> value,
        bool isNew) {
    }
};

int main() {
    system("PAUSE");
    NetworkTable::SetServerMode();
    NetworkTable::Initialize();

    std::shared_ptr<NetworkTable> ptr = NetworkTable::GetTable("SmartDashboard");
    BasicListener* bl = new BasicListener();

    for (int i = 0; i < 1000000; i++) {     
        ptr->AddTableListener(bl);
        ptr->RemoveTableListener(bl);
    }

    NetworkTable::Shutdown();
    delete bl;

    system("PAUSE");
    return 0;
}
```

At the second pause, memory usage is over 100MB.

I did some digging into this issue. Looking at `Notifier::RemoveEntryListener`, it doesn't ever delete from `m_entry_listeners`

From first glance it looks like "uids" are dependant on the ordering of the vector and that's why entry listeners are never deleted. Perhaps `m_entry_listeners` would be better as a `std::unordered_map` or `std::map`

PeterJohnson:
This avoids endless growth of the vector.

Fixes #80.


calcmogul:
In my opinion, the UidVector class could use some comments describing generally what emplace_back() and erase() are doing to the free list. A blurb above each function would suffice.

PeterJohnson:
@calcmogul: Added comments.

calcmogul:
LGTM

PR #84, opened 2016-07-27
Create dummy wpiutil library.

PeterJohnson:
This will allow dependencies such as wpilibc to update to use wpiutil
without breaking "normal" ntcore static library use in the meantime.

This commit also restructures the gradle files by creating a new
(placeholder) wpiutil project, and moving the ntcore project into
a separate gradle file.  Added toolchains/native.gradle (refactored from
ntcore).

Also fixes ntcore skipJava on Windows by providing an alternate .def file
for this case.


ntcore.gradle
```diff
@@ -0,0 +1,166 @@
...
+        group = 'WPILib'
+        destinationDir = project.buildDir
+        baseName = 'ntcore'
+        classifier = "${project.buildPlatform}"
+
+        from(file('include')) {
+            into 'include'
+        }
+
+        if (!hasProperty('skipJava')) {
```
ThadHouse: This needs to be `if (!project.hasProperty('skipJava')) {` to work properly.

PR #86, opened 2016-08-04
Adds timeout capabilities to blocking Rpc Calls

ThadHouse:


include/ntcore_c.h
```diff
@@ -305,6 +305,8 @@ int NT_NotifierDestroyed();
 /*
  * Remote Procedure Call Functions
  */
+ 
+#define TIMEOUT_INDEFINITE -1
```
PeterJohnson: This needs an NT prefix.  Is this really necessary given the non-timeout variants?

ThadHouse: Yeah its probably not actually needed. I'll remove it.

src/Storage.cpp
```diff
@@ -1407,15 +1407,32 @@ unsigned int Storage::CallRpc(StringRef name, StringRef params) {
...
     auto i =
         m_rpc_results.find(std::make_pair(call_uid >> 16, call_uid & 0xffff));
     if (i == m_rpc_results.end()) {
       if (!blocking || m_terminating) return false;
-      m_rpc_results_cond.wait(lock);
+      if (time_out == kTimeout_Indefinite) {
+        m_rpc_results_cond.wait(lock);
+      } else {
+        auto time_out_millis = 
+          std::chrono::milliseconds(static_cast<long long>(time_out * 1000));
```
PeterJohnson: Why do this instead of just converting directly to seconds?

src/RpcServer.h
```diff
@@ -9,6 +9,7 @@
 #define NT_RPCSERVER_H_
 
 #include <atomic>
+#include <chrono>
```
PeterJohnson: Shouldn't need chrono here.. should be in the C++ file instead?

src/Storage.h
```diff
@@ -9,6 +9,7 @@
 #define NT_STORAGE_H_
 
 #include <atomic>
+#include <chrono>
```
PeterJohnson: Shouldn't need chrono here.

src/ntcore_c.cpp
```diff
@@ -8,6 +8,7 @@
 #include "ntcore.h"
 
 #include <cassert>
+#include <chrono>
```
PeterJohnson: Shouldn't need chrono here.

src/RpcServer.cpp
```diff
@@ -61,10 +61,26 @@ void RpcServer::ProcessRpc(StringRef name, std::shared_ptr<Message> msg,
...
 bool RpcServer::PollRpc(bool blocking, RpcCallInfo* call_info) {
+  return PollRpc(blocking, kTimeout_Indefinite, call_info);
+}
+
+bool RpcServer::PollRpc(bool blocking, double time_out, RpcCallInfo* call_info) {
   std::unique_lock<std::mutex> lock(m_mutex);
   while (m_poll_queue.empty()) {
     if (!blocking || m_terminating) return false;
-    m_poll_cond.wait(lock);
+    if (time_out == kTimeout_Indefinite) {
```
PeterJohnson: Should this be < 0 instead of == kTimeout_Indefinite?

include/ntcore_cpp.h
```diff
@@ -9,6 +9,7 @@
 #define NTCORE_CPP_H_
 
 #include <cassert>
+#include <chrono>
```
PeterJohnson: Don't need chrono here.

src/RpcServer.h
ThadHouse: These chronos where an initial version where the functions took a std::milliseconds value. I'll fix them.

ntcore.def
```diff
@@ -90,3 +90,5 @@ NT_SetDefaultEntryRaw @91
 NT_SetDefaultEntryBooleanArray @92
 NT_SetDefaultEntryDoubleArray @93
 NT_SetDefaultEntryStringArray @94
+NT_PollRpcTimeout @95
+NT_GetRpcResultTimeout @96
```
PeterJohnson: No EOL at EOF..

ThadHouse:
Its added. We should make the formatting tools work over here at some point.

PR #87, opened 2016-08-06
Changes remote_name to remote_ip

ThadHouse:
Was confusing, as remote_name seemed like a name but wasn't.
Also changes remote_ip in the C api to use an NT_String

### DIFF
--- a/include/ntcore_c.h
+++ b/include/ntcore_c.h
@@ -120,7 +120,7 @@ struct NT_EntryInfo {
 /** NetworkTables Connection Information */
 struct NT_ConnectionInfo {
   struct NT_String remote_id;
-  char *remote_name;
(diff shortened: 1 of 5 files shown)


333fred:
As much as I like the idea of cleaning up the parameter, this is a breaking change, yes? Do we really want to break the api for just this?

ThadHouse:
This is a breaking change, but I don't think it is going to affect many, if any teams. The old name was very confusing anyway, and I have a feeling we would have gotten a question or 2 on it if a team used it, as it was only added this year. Plus if any teams did, the error message for this is pretty easy to read.

As for the char\* to NT_String change, I would be highly surprised if any team used the raw C api instead of the C++ api, and that change only effects the C api.

PR #89, opened 2016-08-13
Reduced minimum update rate from 100 ms to 10 ms.

PeterJohnson:


ThadHouse:
LGTM

PeterJohnson:
@Kevin-OConnor ok with this change?  Note the default update rate is still 100 ms.

Kevin-OConnor:
LGTM. 

I think the default rate is fine for the vast majority of teams. Lowering the min (and at some point documenting, probably in Screensteps somewhere, that you can speed up the update rate) should cover most of the cases where someone rolled their own protocol just due to speed/latency.

ThadHouse:
i'd probably recommend on screen steps keeping update rate at 100ms anyway, and just have teams flush when they have data they wanted updated faster. That'd probably work better for teams that knew enough and actually needed faster data.

PR #96, opened 2016-08-17
Fixes server side polled rpcs

ThadHouse:


src/RpcServer.cpp
```diff
@@ -84,7 +84,12 @@ bool RpcServer::PollRpc(bool blocking, double time_out, RpcCallInfo* call_info)
   }
 
   auto& item = m_poll_queue.front();
-  unsigned int call_uid = (item.conn_id << 16) | item.msg->seq_num_uid();
+  unsigned int call_uid;
+  // do not include conn id if the result came from the server
+  if (item.conn_id != 0xffff)
+    call_uid = (item.conn_id << 16) | item.msg->seq_num_uid();
+  else
+    call_uid = item.msg->seq_num_uid();
```
PeterJohnson, line 92: This changes both call_info->call_uid and the response map call_uid.  Do both of these need to be changed?  Also, this makes conn_id=0 and conn-id=0xffff equivalent, does this cause any issues anywhere else?

ThadHouse, line 92: Yes, both the call info and the response map need to be equivalent. CallRpc was handing out an id that expected the conn_id in the response map to be 0 (if both called from server side).

This does not cause any issues with 0 and 0xfff being the same, as the server call is handled with a different callback then remote calls.

PR #101, opened 2016-08-21
Updates gradle to 3.0.

333fred:
Now that the check bug is fixed, removes the empty check tasks, which is what I thought would cause us to not be able to use 3.0.


JLLeitschuh:
LGTM!

PR #107, opened 2016-08-29
Switches AppVeyor to use same build for 32 and 64 bit java tests

ThadHouse:


JLLeitschuh:
What failed here?

ThadHouse:
It looks like it just froze. I've seen that happen before randomly. I'll rebase and let it test again.

ThadHouse:
So it does look like it was just a random crash. This definitely would be a nice fix, because right now AppVeyor runs 4 builds

JLLeitschuh:
LGTM!

PR #108, opened 2016-08-30
Skips building a branch when a PR exists

ThadHouse:


ThadHouse:
This doesn't actually work, which means you have to specifically specify branches to build. Closing.

PR #111, opened 2016-09-06
Adds connection listeners that can be called statically

ThadHouse:
This way you can have a connection listener before initializing NetworkTables

### DIFF
--- a/java/src/edu/wpi/first/wpilibj/networktables/NetworkTable.java
+++ b/java/src/edu/wpi/first/wpilibj/networktables/NetworkTable.java
@@ -223,11 +223,11 @@ public class NetworkTable implements ITable, IRemote {
     return !client;
   }
 
-  private class ListenerBase {
+  private static class ListenerBase {
     public int uid;
   }
 
-  private class ConnectionListenerAdapter extends ListenerBase implements NetworkTablesJNI.ConnectionListenerFunction {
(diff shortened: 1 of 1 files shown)


PeterJohnson:
Would C++ benefit from this as well (for API similarity)?  Is this a breaking change?

ThadHouse:
C++ does not have connection listeners at all in the NetworkTables API. And this should not be a breaking change. All of the IRemote functions are valid, and its always been up to the user to know exactly which instance the functions are being called on if the IRemote is received from the callbacks.

java/src/edu/wpi/first/wpilibj/networktables/NetworkTable.java
```diff
@@ -243,6 +243,41 @@ public void apply(int uid, boolean connected, ConnectionInfo conn) {
...
+      return conns.length > 0;
+    }
+    public boolean isServer() {
+      return !client;
+    }
+  };
+  
+  private static final Hashtable<IRemoteConnectionListener,ConnectionListenerAdapter> globalConnectionListenerMap = new Hashtable<IRemoteConnectionListener,ConnectionListenerAdapter>();
+  public static synchronized void addGlobalConnectionListener(IRemoteConnectionListener listener,
+                                                       boolean immediateNotify) {
```
PeterJohnson: Indentation?

ThadHouse: Fixed

PR #112, opened 2016-09-17
Passes the ConnectionInfo of the Rpc client on server callback

ThadHouse:
It's a breaking change to the Rpc callbacks, however I think as long as this happens this fall that is OK, as we've done a few other breaking changes, and I don't think anyone really has used RPCs yet much.


PeterJohnson:
You need to update `NT_DisposeRpcCallInfo` to call `NT_DisposeConnectionInfo` to avoid a memory leak.

ThadHouse:
Done

src/ntcore_c.cpp
```diff
@@ -234,10 +235,14 @@ void NT_CreateRpc(const char *name, size_t name_len, const char *def,
...
+      [=](StringRef name, StringRef params, 
+          const ConnectionInfo& conn_info) -> std::string {
+        NT_ConnectionInfo conn_c;
+        ConvertToC(conn_info, &conn_c);
         size_t results_len;
         char* results_c = callback(data, name.data(), name.size(),
-                                   params.data(), params.size(), &results_len);
+                                   params.data(), params.size(), 
+                                   &conn_c, &results_len);
         std::string results(results_c, results_len);
```
PeterJohnson, line 246: Need to dispose of the NT_ConnectionInfo here.

ThadHouse: Fixed

PeterJohnson:
Update JNI?

ThadHouse:
These are not in JNI yet.

PR #113, opened 2016-09-17
Fully asigns the ConnectionInfo struct

ThadHouse:
For some reason, it was putting garbage in for the things that shouldn't
have been, which doesn't make much sense.


src/Storage.cpp
```diff
@@ -349,6 +349,13 @@ void Storage::ProcessIncoming(std::shared_ptr<Message> msg,
       ConnectionInfo conn_info;
       auto c = conn_weak.lock();
       if (c) conn_info = c->info();
```
PeterJohnson: Should put this on the next line since there's an else

ThadHouse: Done

PR #116, opened 2016-09-19
Adds ConnectionInfo to Rpc callback

ThadHouse:
Only did the callback this time instead of the RpcCallInfo. I'll do that in a separate PR. This has been tested on both linux and windows to work properly

### DIFF
--- a/src/RpcServer.cpp
+++ b/src/RpcServer.cpp
@@ -47,15 +47,17 @@ void RpcServer::Stop() { m_owner.Stop(); }
 
 void RpcServer::ProcessRpc(StringRef name, std::shared_ptr<Message> msg,
                            RpcCallback func, unsigned int conn_id,
-                           SendMsgFunc send_response) {
+                           SendMsgFunc send_response, 
+                           const ConnectionInfo& conn_info) {
   if (func) {
     auto thr = m_owner.GetThread();
     if (!thr) return;
-    thr->m_call_queue.emplace(name, msg, func, conn_id, send_response);
+    thr->m_call_queue.emplace(name, msg, func, conn_id, send_response, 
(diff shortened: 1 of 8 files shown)


src/RpcServer.cpp
```diff
@@ -73,8 +75,8 @@ bool RpcServer::PollRpc(bool blocking, double time_out, RpcCallInfo* call_info)
     } else {
 #if defined(_MSC_VER) && _MSC_VER < 1900
       auto timeout_time = std::chrono::steady_clock::now() + 
-          std::chrono::duration<int64_t, std::nano>(static_cast<int64_t>
-          (time_out * 1e9));
+           std::chrono::duration<int64_t, std::nano>(static_cast<int64_t>
+           (time_out * 1e9));
```
PeterJohnson: 5 spaces?

ThadHouse: Oh whoops. Fixed, and also rebased off of the newest master

src/Storage.cpp
```diff
@@ -1392,13 +1404,19 @@ unsigned int Storage::CallRpc(StringRef name, StringRef params) {
     // gracefully anyway.
     auto rpc_callback = entry->rpc_callback;
     lock.unlock();
+    ConnectionInfo conn_info;
+    conn_info.remote_id = "Server";
+    conn_info.remote_ip = "LocalHost";
```
PeterJohnson: Usually this would be all lowercase "localhost" or "127.0.0.1"?

ThadHouse: Ah I didn't know. I'll fix that.

Fixed

PeterJohnson (approved):

PR #123, opened 2016-10-15
Updated the gradle build to get the version from the WPILibVersioning plugin

333fred:
This is the work to implement wpilibsuite/design-docs#5. This will fail builds until we push an annotated tag. @PeterJohnson, I believe we should start with v3.1.0.


PeterJohnson (approved):

333fred:
Alright, I have no idea what appveyor is doing. It seems like it built, then built again, then built again, and kept building until it timed out. @JLLeitschuh, any ideas?

333fred:
It built twice in total, but successfully, and then got most of the way through a third build.

PR #125, opened 2016-10-17
Optimizes NetworkTables string concatination

ThadHouse:
During testing, I was seeing a lot of unnecessary code (and allocations
in Java/C#) when appending the path separator to the base path. That
technically is a constant, so this constant computes those upon class
create. Saves GC in Java, and some code in C++

### DIFF
--- a/java/src/edu/wpi/first/wpilibj/networktables/NetworkTable.java
+++ b/java/src/edu/wpi/first/wpilibj/networktables/NetworkTable.java
@@ -204,9 +204,11 @@ public class NetworkTable implements ITable, IRemote {
   }
 
   private final String path;
+  private final String pathWithSep;
 
   NetworkTable(String path) {
     this.path = path;
+    this.pathWithSep = path + PATH_SEPARATOR;
   }
   public String toString() { return "NetworkTable: " + path; }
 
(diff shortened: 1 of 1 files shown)


PeterJohnson:
This really shouldn't make a difference in C++... it's all on the stack and just a few extra instructions.  But I definitely see the savings for Java.

ThadHouse:
That's a good point. SmallString has an operator to append just a single char, so it doesn't have to allocate a new string. I can revert the C++ changes. But the Java ones for sure are helpful, since PATH_SEPARATOR is a char, it actually has to get boxed before it gets concatenated. So the previous way resulted in a boxing, and 2 extra string objects being created, every single nt call.

PeterJohnson (approved):

java/src/edu/wpi/first/wpilibj/networktables/NetworkTable.java
```diff
@@ -204,9 +204,11 @@ public synchronized static NetworkTable getTable(String key) {
   }
 
   private final String path;
+  private final String pathWithSep;
 
   NetworkTable(String path) {
     this.path = path;
+    this.pathWithSep = path + PATH_SEPARATOR;
```
JLLeitschuh, line 211: Is the `path` variable still used now given this change?

ThadHouse: Yes there are some places that still need the raw path variable without a postpended separator (Table listeners don't need the sep)

PR #128, opened 2016-10-18
Ignores KeepAlive messages during initial handshakes

### LINKED ISSUE #126: Keep alive message terminates handshake
While you probably shouldn't receive a keepalive at that time, there's no good reason why you shouldn't be able to receive one.

ThadHouse:
Shouldn't happen normally, but could happen if handshakes take a longer
then normal period of time.
Closes #126

### DIFF
--- a/src/Dispatcher.cpp
+++ b/src/Dispatcher.cpp
@@ -403,6 +403,11 @@ bool DispatcherBase::ClientHandshake(
     DEBUG4("received init str=" << msg->str() << " id=" << msg->id()
                                 << " seq_num=" << msg->seq_num_uid());
     if (msg->Is(Message::kServerHelloDone)) break;
+    // shouldn't receive a keep alive, but handle gracefully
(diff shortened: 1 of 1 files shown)


PeterJohnson (approved):

src/Dispatcher.cpp
```diff
@@ -403,6 +403,8 @@ bool DispatcherBase::ClientHandshake(
     DEBUG4("received init str=" << msg->str() << " id=" << msg->id()
                                 << " seq_num=" << msg->seq_num_uid());
     if (msg->Is(Message::kServerHelloDone)) break;
+    // shouldn't receive a keep alive, but handle gracefully
+    if (msg->Is(Message::kKeepAlive)) continue;
```
virtuald: Infinite loop, a new message doesn't get retrieved.

ThadHouse: Fixed

PR #129, opened 2016-10-20
Java: Allow any Number type other than Long to be passed to putValue().

### LINKED ISSUE #72: Add support for adding any number value
Right now, calls to `putValue` with a float value (and I suspect `Integer` as well) cause an `IllegalArgumentException` to be thrown.

Lines 722 and 723 in `NetworkTable.putValue(String, Object)` could be changed to the following bolded lines.

<pre>
  @Override
  public boolean putValue(String key, Object value) throws IllegalArgumentException {
    if (value instanceof Boolean)
      return NetworkTablesJNI.putBoolean(path + PATH_SEPARATOR + key, ((Boolean)value).booleanValue());
    <b>else if (value instanceof Number)
      return NetworkTablesJNI.putDouble(path + PATH_SEPARATOR + key, ((Number)value).doubleValue());</b>
    ... // Ommitted for brevity
</pre>

Similarly, `putNumber(String, double)` could be overloaded with a `putNumber(String, Number)` that would just delegate to the first method:

```
public void putNumber(String key, Number value) {
  putNumber(key, value.doubleValue());
}
```

PeterJohnson:
Also improve exception message.

Fixes #72.


java/src/edu/wpi/first/wpilibj/networktables/NetworkTable.java
```diff
@@ -850,7 +850,7 @@ public boolean putRaw(String key, ByteBuffer value, int len) {
   public boolean putValue(String key, Object value) throws IllegalArgumentException {
     if (value instanceof Boolean)
       return NetworkTablesJNI.putBoolean(path + PATH_SEPARATOR + key, ((Boolean)value).booleanValue());
-    else if (value instanceof Double)
+    else if (value instanceof Number && !(value instanceof Long))
       return NetworkTablesJNI.putDouble(path + PATH_SEPARATOR + key, ((Double)value).doubleValue());
```
ThadHouse: Can Number be cast to a Double? Or does this need to be cast to Number then the double taken from there?
EDIT: Looking at the docs not all numbers are double, so we do need to cast to Number then call doubleValue.

PeterJohnson: Fixed.

java/src/edu/wpi/first/wpilibj/networktables/NetworkTable.java
```diff
@@ -852,8 +852,8 @@ public boolean putRaw(String key, ByteBuffer value, int len) {
   public boolean putValue(String key, Object value) throws IllegalArgumentException {
     if (value instanceof Boolean)
       return NetworkTablesJNI.putBoolean(pathWithSep + key, ((Boolean)value).booleanValue());
-    else if (value instanceof Double)
-      return NetworkTablesJNI.putDouble(pathWithSep + key, ((Double)value).doubleValue());
+    else if (value instanceof Number && !(value instanceof Long))
```
333fred: Not sure if this is the way we want to do this. BigDecimal, for example, is a Number. If we're excluding Long, we should exclude everything that doesn't fit. Doing this with casts seems fundamentally broken in terms of object oriented languages. Number has a `.doubleValue()`. We should just call that, and document that rounding can/will occur as Numbers are sent as doubles.

333fred (changes requested):

java/src/edu/wpi/first/wpilibj/networktables/NetworkTable.java
PeterJohnson: We are calling Number.doubleValue() now.  Are you recommending we have a tighter list of allowed types rather than assuming Number is okay and only excluding the ones we know don't work?

AustinShalit: It is future-proof to whitelist rather than blacklist.

333fred: I'm not sure which is correct, but we need to either create a whitelist or allow everything in. And I  _really_ don't like the idea of a whitelist.

PeterJohnson: We could just allow everything in and say caveat emptor.

333fred: I think that's what I'm in favor of.

PeterJohnson: Done

java/src/edu/wpi/first/wpilibj/networktables/NetworkTable.java
```diff
@@ -864,8 +864,8 @@ else if (value instanceof double[])
       return NetworkTablesJNI.putDoubleArray(pathWithSep + key, (double[])value);
     else if (value instanceof Boolean[])
       return NetworkTablesJNI.putBooleanArray(pathWithSep + key, toNative((Boolean[])value));
-    else if (value instanceof Double[])
-      return NetworkTablesJNI.putDoubleArray(pathWithSep + key, toNative((Double[])value));
+    else if (value instanceof Number[])
+      return NetworkTablesJNI.putDoubleArray(pathWithSep + key, toNative((Number[])value));
```
nightpool, line 868: we're not using doubleValue here—looks like this isn't doing what it should do. If we want to do the same thing, we probably have to construct a new array by calling doubleValue on each member of `value`, and pass that. #orsomething

PeterJohnson: We do exactly that... take a look at the implementation of (and the changes in this PR to) `toNative()`.

333fred (approved):

java/src/edu/wpi/first/wpilibj/networktables/NetworkTable.java
nightpool: Oops. My bad. I totally missed that they were the same toNative.

java/src/edu/wpi/first/wpilibj/networktables/NetworkTable.java
```diff
@@ -864,8 +864,8 @@ else if (value instanceof double[])
       return NetworkTablesJNI.putDoubleArray(pathWithSep + key, (double[])value);
     else if (value instanceof Boolean[])
       return NetworkTablesJNI.putBooleanArray(pathWithSep + key, toNative((Boolean[])value));
-    else if (value instanceof Double[])
-      return NetworkTablesJNI.putDoubleArray(pathWithSep + key, toNative((Double[])value));
+    else if (value instanceof Number[])
```
SamCarlberg, line 867: This only works for boxed number arrays (`Integer[]`, `Float[]`, `Double[]`, etc) and fails for the more commonly used primitive arrays (`int[]`, `float[]`, `double[]`, ...)

333fred: That should be fine. They should be auto-boxed, as this function takes an object.

SamCarlberg: All arrays subclass `Object`, so there's no autoboxing.

Take this simple method:

``` java
public static boolean isNumberArray(Object arr) {
  return arr instanceof Number[];
}
```

This will return `true` if passed an array of any _boxed_ number type, and `false` for everything else

PR #143, opened 2016-10-25
Client: when reconnecting, default to NT3.

### LINKED ISSUE #142: Cannot switch back to NT3 mode after connecting to NT2 server
See dispatcher implementation, m_reconnect_proto_rev never gets set back to 0x0300.

PeterJohnson:
Previously once the client fell back once to NT2, it would never try
connecting as NT3 even if the server was replaced with a NT3-capable
one.

Fixes #142.


ThadHouse (approved):
LGTM!

333fred (approved):

PR #151, opened 2016-11-05
Add FRC Driver Station connection support.

PeterJohnson:
The 2017 FRC Driver Station supports getting the robot IP via a TCP
connection that returns JSON.  Use this to support overriding the
server IP address used for client connections.

Default to using this approach for client connections in the NetworkTable
interfaces.

Also add support for setting the server address without stopping and
restarting the client.


include/ntcore_c.h
```diff
@@ -363,6 +363,11 @@ void NT_StartServer(const char* persist_filename, const char* listen_address,
 void NT_StopServer(void);
 
 /** Starts Client
+ * Starts a client.  Use NT_SetServer to set the server name and port.
+ */
+void NT_StartClientNone(void);
```
virtuald, line 368: In pynetworktables I have a test mode that probably does the same thing as this

src/networktables/NetworkTable.cpp
```diff
@@ -44,23 +41,41 @@ void NetworkTable::SetClientMode() { s_client = true; }
...
-#ifdef _MSC_VER
-  sprintf_s(tmp, "roboRIO-%d-FRC.local", team);
-#else
-  using namespace std;
-  snprintf(tmp, 30, "roboRIO-%d-FRC.local", team);
-#endif
-  SetIPAddress(tmp);
+  std::pair<StringRef, unsigned int> servers[3];
+
+  // roboRIO-<team>-FRC.local
```
virtuald: Is there a preferred ordering or are they done in parallel? If ordered, I'd prefer the IP address to be first.

src/networktables/NetworkTable.cpp
```diff
@@ -20,11 +20,8 @@ unsigned int NetworkTable::s_port = NT_DEFAULT_PORT;
 void NetworkTable::Initialize() {
   if (s_running) Shutdown();
   if (s_client) {
-    std::vector<std::pair<StringRef, unsigned int>> servers;
-    servers.reserve(s_ip_addresses.size());
-    for (const auto& ip_address : s_ip_addresses)
-      servers.emplace_back(std::make_pair(ip_address, s_port));
-    nt::StartClient(servers);
+    nt::StartClient();
+    nt::SetServerFromDS(s_port);
```
virtuald: Should this be done automatically? It feels like it's something that should be enabled explicitly?

src/Dispatcher.cpp
```diff
@@ -195,6 +199,27 @@ void DispatcherBase::NotifyConnections(
...
+  connectors.push_back(connector);
+  SetConnector(std::move(connectors));
+}
+
+void DispatcherBase::SetConnector(std::vector<Connector>&& connectors) {
+  std::lock_guard<std::mutex> lock(m_user_mutex);
+  m_client_connectors = std::move(connectors);
+}
+
+void DispatcherBase::SetConnectorOverride(Connector connector) {
```
virtuald, line 213: This connector override thing seems strange to me. Why not just query all of the possible connectors for potential addresses to connect to?

src/networktables/NetworkTable.cpp
PeterJohnson: It's round robin, so all will get tried.

src/Dispatcher.cpp
PeterJohnson: The idea is for the DS discovered IP to not completely clobber the user-set server list, so when the DS disconnects, we fall back to trying the user list.

include/ntcore_c.h
PeterJohnson: This really isn't a test mode.  It's just allowing separation of starting up the client and setting the server addresses.

PR #155, opened 2016-11-15
Adds a c_str method to StringRef

ThadHouse:
Current class does not ensure the string is null terminated. As there is
no defined error to actually check for a null terminator, we determine
if it is null terminated based on the constructor type. Then if on c_str
it is not known to be null terminated, we use a passed in buffer to copy
the string and ensure null termination

### DIFF
--- a/wpiutil/include/llvm/StringRef.h
+++ b/wpiutil/include/llvm/StringRef.h
@@ -51,6 +51,8 @@ namespace llvm {
     const char *Data;
 
     /// The length of the string.
+    /// MSB of length indicates if we are null terminated or not
+    /// Flag set is null terminated, flag not set is not
     size_t Length;
 
     // Workaround memcmp issue with null pointers (undefined behavior)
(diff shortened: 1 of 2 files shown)


wpiutil/src/llvm/StringRef.cpp
```diff
@@ -120,6 +120,21 @@ std::string StringRef::upper() const {
...
+  if (is_null_terminated()) {
+    // If null terminated, return data directly
+    return data();
+  } else {
+    // If not null terminated, use SmallVectorImpl to store data
+    // copy data, and return a known null terminated string
+    buf.clear();
+    buf.append(begin(), end());
+    buf.push_back(0);
+    buf.pop_back();
```
PeterJohnson: Not necessary to call `pop_back()`.

wpiutil/include/llvm/StringRef.h
```diff
@@ -72,26 +83,39 @@ namespace llvm {
...
 
     /// Construct a string ref from an std::string.
     /*implicit*/ StringRef(const std::string &Str)
-      : Data(Str.data()), Length(Str.length()) {}
+      : Data(Str.data()), Length(Str.length()) {
+        // Require length to not use MSB of size
+        assert(Length < ~(std::numeric_limits<size_t>::max() << 1));
+        // If from a std::string, we are null terminated
+        set_null_terminated(true);
+      }
```
PeterJohnson: Indentation...

PeterJohnson (approved):

PR #159, opened 2016-11-25
Fixed unused variable warning from gcc.

AustinSchuh:
This makes it so that ntcore builds with -Wextra -Werror

### DIFF
--- a/wpiutil/src/llvm/raw_ostream.cpp
+++ b/wpiutil/src/llvm/raw_ostream.cpp
@@ -845,12 +845,11 @@ raw_null_ostream::~raw_null_ostream() {
 #endif
 }
 
-void raw_null_ostream::write_impl(const char *Ptr, size_t Size) {
-}
+void raw_null_ostream::write_impl(const char * /*Ptr*/, size_t /*Size*/) {}
 
 uint64_t raw_null_ostream::current_pos() const {
   return 0;
 }
 
-void raw_null_ostream::pwrite_impl(const char *Ptr, size_t Size,
-                                   uint64_t Offset) {}
+void raw_null_ostream::pwrite_impl(const char * /*Ptr*/, size_t /*Size*/,
+                                   uint64_t /*Offset*/) {}


PeterJohnson:
@frcjenkins test this please

wpiutil/src/llvm/raw_ostream.cpp
```diff
@@ -845,12 +845,12 @@ raw_null_ostream::~raw_null_ostream() {
 #endif
 }
 
-void raw_null_ostream::write_impl(const char *Ptr, size_t Size) {
+void raw_null_ostream::write_impl(const char * /*Ptr*/, size_t /*Size*/) {
```
calcmogul: I prefer static casting unused arguments to void to fix this warning. It keeps the function signatures clean.

AustinSchuh: Sure, done.

https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions suggests /**/, but I'm happy to do whatever you guys want.

calcmogul: I did some testing. By commenting out the variable name, maintainers can be sure that variable is unused throughout the whole function. If a cast to void is used instead, there is no such guarantee because maintainers can use the variable without removing the void cast.

Given that, I now prefer what the style guide mandates. What do you think, @PeterJohnson?

AustinSchuh: When you two come to consensus, I'll push out an updated patch.

calcmogul: Peter on Slack: ```it looks like llvm prevailing style is to silence variables only used in asserts with `(void)`, but silence args with `/*arg*/`.  So let's go with the latter since it's both an LLVM file and consistent with our style guide.```

AustinSchuh: Done.

PeterJohnson:
@frcjenkins retest this please

PeterJohnson (approved):

PR #161, opened 2016-11-26
NetworkStream: Add setBlocking() and getNativeHandle().

PeterJohnson:
Also add checking for would block errors in send() and receive().


333fred:
Why are we ignoring the output of fnctl/ioctlsocket?

PeterJohnson:
Changed to pay attention to the return value in both TCPStream and TCPConnector (where it wasn't being checked previously)

333fred (approved):

PR #163, opened 2016-12-15
Add null checks on JNI parameters

PeterJohnson:
Also a number of jni_util additions and cleanups.

### DIFF
--- a/java/lib/NetworkTablesJNI.cpp
+++ b/java/lib/NetworkTablesJNI.cpp
@@ -31,6 +31,7 @@ static JClass entryInfoCls;
 static JClass keyNotDefinedEx;
 static JClass persistentEx;
 static JClass illegalArgEx;
+static JClass nullPointerEx;
 // Thread-attached environment for listener callbacks.
 static JNIEnv *listenerEnv = nullptr;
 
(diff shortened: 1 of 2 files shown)


java/lib/NetworkTablesJNI.cpp
```diff
@@ -773,6 +958,14 @@ JNIEXPORT jdoubleArray JNICALL Java_edu_wpi_first_wpilibj_networktables_NetworkT
 JNIEXPORT jobjectArray JNICALL Java_edu_wpi_first_wpilibj_networktables_NetworkTablesJNI_getStringArray__Ljava_lang_String_2_3Ljava_lang_String_2
   (JNIEnv *env, jclass, jstring key, jobjectArray defaultValue)
 {
+  if (!key) {
+    env->ThrowNew(nullPointerEx, "key cannot be null");
+    return nullptr;
+  }
+  if (!defaultValue) {
```
ThadHouse: I don't think we should throw on a null defaultValue. I know teams in the past have used null as their defaultValue since it's one of the cases that doesn't often get returned from ntcore, so its easy to check.

ThadHouse (changes requested):

java/lib/NetworkTablesJNI.cpp
PeterJohnson: fixed

ThadHouse (approved):
LGTM

PR #165, opened 2016-12-19
Add new ARM Toolchain, add toolchainpath to README

JaciBrunning:
This Pull Request is nearly identical to [allwpilib#405](https://github.com/wpilibsuite/allwpilib/pull/405)

In short, currently a hack exists in nt-core and many other WPILib Suite projects for the ARM / RoboRIO Cross Toolchain, specifically for Mac platforms. Gradle detects Mac GCC as a proxy to XCode Clang, so it sees it as not buildable unless a Clang target is also registered. The detection for this proxy is based on the output of `gcc`, not `g++`. As NT-Core does not declare a C Compiler Tool, but instead a C++ Compiler Tool, it defaults to system gcc. This PR fixes this hack and correctly identifies it as GCC and not Clang on OS X.

Furthermore, the `toolChainPath` property has been added to the project that allows the user to specify the Toolchain Path should it not be on the System PATH.

This PR is also partial preparation for the build system overhaul discussed [here](https://github.com/wpilibsuite/allwpilib/pull/405#issuecomment-267867080) for later in 2017 should the interest be there.

This PR has been tested on OS X El Capitan 10.11.6

### DIFF
--- a/toolchains/arm.gradle
+++ b/toolchains/arm.gradle
@@ -2,18 +2,21 @@ ext.isArm = true
 ext.buildPlatform = 'arm'
 
 def compilerPrefix = project.hasProperty('compilerPrefix') ? project.compilerPrefix : 'arm-frc-linux-gnueabi-'
+def toolChainPath = project.hasProperty('toolChainPath') ? project.toolChainPath : null
 model {
     platforms {
-	arm {
-	    architecture 'arm'
-	    operatingSystem 'linux'
-	}
+        arm {
+            architecture 'arm'
+            operatingSystem 'linux'
+        }
     }
     toolChains {
-        gcc(Gcc) {
+        armGcc(Gcc) {
+            if (toolChainPath != null) path(toolChainPath)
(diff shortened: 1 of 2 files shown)


PeterJohnson:
@frcjenkins test this please

toolchains/arm.gradle
```diff
@@ -2,18 +2,21 @@ ext.isArm = true
...
-	    operatingSystem 'linux'
-	}
+        arm {
+            architecture 'arm'
+            operatingSystem 'linux'
+        }
     }
     toolChains {
-        gcc(Gcc) {
+        armGcc(Gcc) {
```
333fred, line 14: Were you going to change this roborioArm as well?

JaciBrunning:
Given that NT-Core can be built on more ARM platforms than just the RoboRIO I decided against it. In the eventual overhaul I plan to include both a RoboRIO toolchain and one for arm in general

333fred (approved):

PR #170, opened 2016-12-25
Use InetNtop on Win32 rather than WSAAddressToString.

PeterJohnson:
Similarly, use InetPton rather than WSAStringToAddress.

The WSAAddressToString function is intended to provide a user-readable
string and thus includes the port number.  This breaks some use cases
on Windows which expect to get just the IP address.

Note: The InetPton and InetNtop functions are available only in Vista or above.


AustinShalit:
closes wpilibsuite/SmartDashboard/pull/57

PR #173, opened 2016-12-26
Removes all TCPAcceptor checks

ThadHouse:
Other fix doesn't fix, and this seems like a deeper issue in JStringRef
that I don't have time to fix. This will work around the issue for now.


PeterJohnson:
Closing in favor of #175 which fixes the underlying issue.

Issue #174, opened 2016-12-26
JStringRef does not handle empty strings properly


ThadHouse:
Instead of returning an empty string ref, it seems to be returning some garbage characters.

PeterJohnson:
Fixed in #175.

PR #175, opened 2016-12-26
JStringRef: ensure string is null terminated so c_str() works correctly.

PeterJohnson:


PeterJohnson:
Fixes #174.

AustinShalit:
This works for me

PR #177, opened 2017-01-03
Adds support for building separate armhf artifacts

ThadHouse:
Currently if using a separate compiler prefix, it would get published to
the arm classifier. This modifies so there is now an armhf classifier
that is used if a separate compiler prefix is used.

### DIFF
--- a/ntcore.gradle
+++ b/ntcore.gradle
@@ -61,7 +61,12 @@ def ntcoreZipTask = { project ->
         group = 'WPILib'
         destinationDir = project.buildDir
         baseName = 'ntcore'
-        classifier = "${project.buildPlatform}"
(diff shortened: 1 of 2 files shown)


333fred (changes requested):
Not a fan of this method of telling the difference. Just because I'm using a compiler prefix does not mean I'm compiling for hard float.

ThadHouse:
I was actually just thinking about that too. I think I'd rather change this to specifically allow another property that would set the postfix. A little more work for the dependencies, but not too bad.

ThadHouse:
@333fred What do you think of the update? It now has a separate property to choose a prefix.

333fred (approved):

PR #178, opened 2017-01-03
Fixes SetIntArrayRegion template for windows builds

ThadHouse:


PeterJohnson:
This actually highlights a potential bug.. should we static_assert sizeof(jint) == sizeof(T) here, and if that's not the case, do an element-by-element assign?  It would require some more template magic..

PeterJohnson (approved):

PR #179, opened 2017-01-04
Correct platform path for Linux on 32-bit Intel

JosephJackson:
Gradle needs to produce a platform path of "Linux/i386" when targeting a Linux 32-bit Intel platform. Otherwise, it doesn't match Java's os.name/os.arch when loading the ntcore library in NetworkTablesJNI.java. Windows uses "x86" but Linux uses "i386". (http://lopica.sourceforge.net/os.html)


333fred:
@frcjenkins ok to test

PeterJohnson (approved):

PR #182, opened 2017-01-05
Adds way to force publishing version from command line

ThadHouse:
In case we need to do a recreation of an artifact for some reason, this
makes it possible.


publish.gradle
```diff
@@ -49,6 +50,18 @@ if (!hasProperty('releaseType')) {
     }
 }
 
+if (project.hasProperty("ntPublishVersion")) {
```
333fred: Could you move this to just under where you define ntVersion? Otherwise, on first glance it looks like a use-before-assignment bug. It's not, but it hurts anyway.

ThadHouse: I can't actually. It breaks the lifecycle.

ThadHouse: Fixed.

333fred (approved):

PR #187, opened 2017-01-27
Adds field IP to round robin list

ThadHouse:
With how many more coprocessors are being added, this is probably needed
to make the field crew sane.

### DIFF
--- a/src/networktables/NetworkTable.cpp
+++ b/src/networktables/NetworkTable.cpp
@@ -43,7 +43,7 @@ void NetworkTable::SetClientMode() { s_client = true; }
 void NetworkTable::SetServerMode() { s_client = false; }
 
 void NetworkTable::SetTeam(int team) {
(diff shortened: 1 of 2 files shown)


src/networktables/NetworkTable.cpp
```diff
@@ -73,6 +73,14 @@ void NetworkTable::SetTeam(int team) {
     servers[3] = std::make_pair(oss.str(), s_port);
   }
 
+  // roboRIO-<team>-FRC.frc-field.local
+  llvm::SmallString<32> field_local;
```
PeterJohnson: This should be bigger than 32 otherwise you end up with an allocation for many team numbers.

ThadHouse: Fixed

PeterJohnson (approved):

PR #200, opened 2017-05-23
Expose ITable paths

### LINKED ISSUE #196: Can't get the path of an ITable
It'd be really nice to have for things that are passed a table and want to find keys under it, sibling tables and keys, or other tables that don't share a namespace on some level.

This applies to both C++ and Java.

SamCarlberg:
Closes #196


PeterJohnson (approved):

ThadHouse:
Is there any way we can have a sensible default for this? Otherwise this is a breaking change. I would be inclined to just return root unless it's specifically implemented, but idk. Or do we care about it breaking

SamCarlberg:
How is it breaking?

ThadHouse:
It adds a new interface method without a default implementation. Technically that's a source and binary breaking change without a default implementation.

PeterJohnson:
Do people actually implement their own ITable's?  I've always wondered why this really needed to be an interface considering the only implementation I'm aware of is NetworkTable...

SamCarlberg:
I'm pretty sure ITable is only implemented by us and mocks for testing (eg GRIP). ITable exists because Mitchell Wills had read a book on how to write modular code and went nuts with interfaces like ITable, IRemote, ITableListener.

ThadHouse:
I used it with implement instance based NetworkTables. However I still think keeping it around is a good idea if we ever do come up with a reason to add a second one. However, I don't think the breaking change would be a big problem.

PR #202, opened 2017-05-28
Add functionailty for getting the current network mode

SamCarlberg:
Proposed modes:

NONE (nothing started or failure)
SERVER (running a server)
CLIENT (running a client)


AustinShalit (commented):
We should update the  isServer and isClient methods in NetworkTables too.

ThadHouse:
You need to add the C function definition and the JNI function definition to the .def files in the root. Both functions must be added to the jni one, and only the C one to the non jni one.

SamCarlberg:
Where would they go? I assume at the end and with an incremented `@` number

ThadHouse:
The C function goes at the bottom of the C declarations with an incremented @ number, and the JNI one goes at the bottom of the JNI functions (Those don't have an ordinal)

src/Dispatcher.cpp
```diff
@@ -83,7 +88,7 @@ void DispatcherBase::StartServer(
     if (m_active) return;
     m_active = true;
   }
-  m_server = true;
+  m_networkMode = NT_NET_MODE_SERVER;
```
ThadHouse: This shouldn't be set here, as this results in a race condition. The mode would say server, even though at this point the server might not have actually started. Put these in the individual threads after Start() (server) and Accept() (client) have successfully finished.

ThadHouse (dismissed):

src/Dispatcher.cpp
SamCarlberg: So in `ServerThreadMain` and `ClientThreadMain`?

ThadHouse: Yeah. I would actually add 2 new states to the enum as well.

NT_NET_MODE_STARTING_SERVER and NT_NET_MODE_STARTING_CLIENT. Set those in StartServer and StartClient, then change them to NT_NET_MODE_SERVER and NT_NET_MODE_CLIENT in `ServerThreadMain` and `ClientThreadMain`

SamCarlberg: Might be worthwhile to have `NT_NET_MODE_SERVER_FAIL` and `NT_NET_MODE_CLIENT_FAIL` as well instead of having `NT_NET_MODE_NONE` be a catchall

ThadHouse: Yeah potentially.

src/Dispatcher.cpp
```diff
@@ -264,10 +269,10 @@ void DispatcherBase::DispatchThreadMain() {
         // post outgoing messages if connection is active
         // only send keep-alives on client
         if (conn->state() == NetworkConnection::kActive)
-          conn->PostOutgoing(!m_server);
+          conn->PostOutgoing(m_networkMode & NT_NET_MODE_CLIENT);
```
ThadHouse: Shouldn't `!m_server` become `!(m_networkMode & NT_NET_MODE_SERVER)`? Otherwise we might have some subtle breaking changes.

SamCarlberg: "only send keep-alives on client" implies to me an explicit check for client-ness

`!(m_networkMode & NT_NET_MODE_SERVER)` will also be true if the current mode is `NT_NET_MODE_NONE`

ThadHouse: Well before that would have been the behavior, since you could have m_server and !m_active. On the topic of that, I'd have to look through the code more, but what about m_active being changed to that flag? that way theyre never out of sync. I'd definitely need @PeterJohnson opinion on that one though.

src/Dispatcher.cpp
```diff
@@ -75,6 +75,11 @@ DispatcherBase::~DispatcherBase() {
   Stop();
 }
 
+unsigned int DispatcherBase::GetNetworkMode() const {
+  return m_networkMode;
+}
+
+
```
PeterJohnson: Extra EOL

src/Dispatcher.cpp
```diff
@@ -243,7 +248,7 @@ void DispatcherBase::DispatchThreadMain() {
     if (!m_active) break;  // in case we were woken up to terminate
 
     // perform periodic persistent save
-    if (m_server && !m_persist_filename.empty() && start > next_save_time) {
+    if (m_networkMode & NT_NET_MODE_SERVER && !m_persist_filename.empty() && start > next_save_time) {
```
PeterJohnson: I prefer `(m_networkMode & NT_NET_MODE_SERVER) != 0` to avoid any potential order-of-operations confusion.

src/Dispatcher.cpp
```diff
@@ -264,10 +269,10 @@ void DispatcherBase::DispatchThreadMain() {
         // post outgoing messages if connection is active
         // only send keep-alives on client
         if (conn->state() == NetworkConnection::kActive)
-          conn->PostOutgoing(!m_server);
+          conn->PostOutgoing(m_networkMode & NT_NET_MODE_CLIENT);
 
         // if client, reconnect if connection died
-        if (!m_server && conn->state() == NetworkConnection::kDead)
+        if (m_networkMode & NT_NET_MODE_CLIENT && conn->state() == NetworkConnection::kDead)
```
PeterJohnson: `(m_networkMode & NT_NET_MODE_CLIENT) != 0`

src/Dispatcher.cpp
```diff
@@ -551,7 +570,7 @@ bool DispatcherBase::ServerHandshake(
 }
 
 void DispatcherBase::ClientReconnect(unsigned int proto_rev) {
-  if (m_server) return;
+  if (m_networkMode & NT_NET_MODE_SERVER) return;
```
PeterJohnson: `(m_networkMode & NT_NET_MODE_SERVER) != 0`

PeterJohnson (approved):

PR #204, opened 2017-06-05
Implement independent instances.

PeterJohnson:
Previously, most of the classes were implemented as singletons so only one
instance was possible.

This change adds an instance handle-based API.

The original API is still provided for backwards compatibility, and simply
are wrapper calls to a "default" instance.


include/ntcore_c.h
```diff
@@ -283,67 +491,107 @@ struct NT_EntryInfo* NT_GetEntryInfo(const char* prefix, size_t prefix_len,
...
+ * @param inst      instance handle
+ */
+void NT_Flush2(NT_Inst inst);
+
 /*
  * Callback Creation Functions
  */
 
 void NT_SetListenerOnStart(void (*on_start)(void* data), void* data);
+void NT_SetListenerOnStart2(NT_Inst inst, void (*on_start)(void* data),
```
ThadHouse: The new listeners that take an instance should also return the instance in the callback, so the instance can be tetermined.

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTableValue.java
```diff
@@ -0,0 +1,325 @@
...
+    m_value = value;
+  }
+
+  NetworkTableValue(int type, long lastChange, Object value) {
+    this(NetworkTableType.getFromInt(type), lastChange, value);
+  }
+
+  NetworkTableValue(NetworkTableType type, Object value) {
+    this(type, NetworkTablesJNI.now(), value);
+  }
```
AustinShalit: Should these constructors be public?

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTableType.java
```diff
@@ -0,0 +1,40 @@
...
+ * Network table data types.
+ */
+public enum NetworkTableType {
+  kUnassigned(0),
+  kBoolean(0x01),
+  kDouble(0x02),
+  kString(0x04),
+  kRaw(0x08),
+  kBooleanArray(0x10),
+  kDoubleArray(0x20),
```
AustinShalit: Should this be NumberArray?

PeterJohnson: And kNumber instead of kDouble?  Probably for consistency, although the C API uses the Double nomenclature instead of Number.  We could easily define both, I guess.

AustinShalit: That might be more confusing.

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTableType.java
```diff
@@ -0,0 +1,40 @@
...
+      case 0x02: return kDouble;
+      case 0x04: return kString;
+      case 0x08: return kRaw;
+      case 0x10: return kBooleanArray;
+      case 0x20: return kDoubleArray;
+      case 0x40: return kStringArray;
+      case 0x80: return kRpc;
+      default: return kUnassigned;
+    }
+  }
```
AustinShalit: If you do this then you are not writing the enum twice and cannot make a typo:
```
    for (NetworkTableType type : NetworkTableType.values()) {
      if (type.value == value) {
        return type;
      }
    }
    return kUnassigned;
```

PeterJohnson: This is not nearly as efficient, namely because values() allocates a new array every time it's called.

src/main/native/cpp/jni/NetworkTablesJNI.cpp
```diff
@@ -99,6 +109,14 @@ JNIEXPORT jint JNICALL JNI_OnLoad(JavaVM *vm, void *reserved) {
   classNotFoundEx = JException(env, "java/lang/ClassNotFoundException");
   if (!classNotFoundEx) return JNI_ERR;
 
+  // Register natives using inner classes (works around JDK-8145897)
+  JClass cls(env, "edu/wpi/first/wpilibj/networktables/NetworkTablesJNI");
```
ThadHouse: This class being cleaned up when OnLoad exits is going to be ok, right? We don't need to hold onto it?

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTableType.java
JLLeitschuh: So store a static copy of the array in a private static variable if you are really concerned about the efficiency?

> Premature optimization is the root of all evil 
> -- [DonaldKnuth](http://wiki.c2.com/?PrematureOptimization)

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTableType.java
```diff
@@ -0,0 +1,40 @@
...
+  kBoolean(0x01),
+  kDouble(0x02),
+  kString(0x04),
+  kRaw(0x08),
+  kBooleanArray(0x10),
+  kDoubleArray(0x20),
+  kStringArray(0x40),
+  kRpc(0x80);
+
+  private int value;
```
JLLeitschuh: Make this final?

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTableInstance.java
```diff
@@ -0,0 +1,421 @@
...
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * NetworkTables Instance
+ */
+public class NetworkTableInstance {
```
JLLeitschuh: Make final?

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTableValue.java
```diff
@@ -0,0 +1,325 @@
+package edu.wpi.first.wpilibj.networktables;
+
+import java.util.Objects;
+
+/**
+ * A network table entry value.
+ */
+public class NetworkTableValue {
+  public NetworkTableValue(NetworkTableType type, long lastChange, Object value) {
```
JLLeitschuh: Do you want to leave these constructors public or only expose the static methods?

Effective Java point 1:
> Item 1: Consider static factory methods instead of constructors
> -- http://mishadoff.com/blog/effective-java/

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTableValue.java
```diff
@@ -0,0 +1,325 @@
+package edu.wpi.first.wpilibj.networktables;
+
+import java.util.Objects;
+
+/**
+ * A network table entry value.
+ */
+public class NetworkTableValue {
```
JLLeitschuh: Do you want to allow subclassing? If not this should probably be final.

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTablesJNI.java
```diff
@@ -54,106 +54,283 @@ else if (System.getProperty("os.name").startsWith("Mac"))
...
   public static native int addEntryListener(String prefix, EntryListenerFunction listener, int flags);
-  public static native void removeEntryListener(int entryListenerUid);
+  public static native int addEntryListener(int inst, String prefix, EntryListenerFunction2 listener, int flags);
+  public static native int addEntryListener(int entry, EntryListenerFunction2 listener, int flags);
+  public static native void removeEntryListener(int entryListener);
 
   @FunctionalInterface
   public interface ConnectionListenerFunction {
-    void apply(int uid, boolean connected, ConnectionInfo conn);
+    public void apply(int connListener, boolean connected, ConnectionInfo conn);
```
JLLeitschuh: Unnecessary access modifier. Public interfaces's methods are public by default.

Also, can you add a javadoc to this method/interface, especially, what is the `connListener` parameter?

src/main/java/edu/wpi/first/wpilibj/networktables/TableEntryListener.java
```diff
@@ -0,0 +1,19 @@
...
+   * Called when a key-value pair is changed in a {@link NetworkTable}.
+   *
+   * @param table the table the key-value pair exists in
+   * @param key the key associated with the value that changed
+   * @param entry the entry associated with the value that changed
+   * @param value the new value
+   * @param flags update flags; for example, EntryListenerFlags.kNew if the key
+   * did not previously exist in the table
+   */
+  public void valueChanged(NetworkTable table, String key, NetworkTableEntry entry, NetworkTableValue value, int flags);
```
JLLeitschuh: Unnecessary `public` keyword.

src/main/java/edu/wpi/first/wpilibj/networktables/TableListener.java
```diff
@@ -0,0 +1,16 @@
...
+@FunctionalInterface
+public interface TableListener {
+  /**
+   * Called when a new table is created within a {@link NetworkTable}.
+   *
+   * @param parent the parent of the table
+   * @param name the name of the new table
+   * @param table the new table
+   */
+  public void tableCreated(NetworkTable parent, String name, NetworkTable table);
```
JLLeitschuh: Unnecessary public keyword.

src/main/native/cpp/networktables/NetworkTableInstance.cpp
```diff
@@ -0,0 +1,62 @@
...
+      },
+      flags);
+}
+
+NT_ConnectionListener NetworkTableInstance::AddConnectionListener(
+    ConnectionListener callback, bool immediate_notify) const {
+  return ::nt::AddConnectionListener(
+      m_handle,
+      [=](NT_ConnectionListener
+          /*uid*/,
```
JLLeitschuh: Was this comment left in the code intentionally?

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTableType.java
PeterJohnson: What's the concern with a switch statement again?  We will basically never change the data type enum values.

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTableValue.java
PeterJohnson: They were private originally.  I'm not sure why @AustinShalit wanted them public.

src/main/native/cpp/networktables/NetworkTableInstance.cpp
PeterJohnson: Yes.  It is a common idiom to comment out the parameter variable name when the parameter is unused in the function, rather than just deleting the parameter name entirely.

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTableValue.java
PeterJohnson: Generally we have not made classes "final" in our APIs.

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTableType.java
ThadHouse: I personally prefer the switch as well. It's more right imo

include/ntcore_c.h
PeterJohnson: The data parameter can be used for that.

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTableValue.java
JLLeitschuh: Just making sure that that's the sort of forwards compatibility guarantee you want to support in your API.

src/main/java/edu/wpi/first/networktables/NetworkTableValue.java
```diff
@@ -0,0 +1,472 @@
...
+    m_time = time;
+  }
+
+  NetworkTableValue(NetworkTableType type, Object value) {
+    this(type, value, NetworkTablesJNI.now());
+  }
+
+  NetworkTableValue(int type, Object value, long time) {
+    this(NetworkTableType.getFromInt(type), value, time);
+  }
```
AustinShalit, line 28: Should these constructors be public?

PeterJohnson: No, because there are factory functions for values.

src/main/java/edu/wpi/first/networktables/ConnectionNotification.java
```diff
@@ -0,0 +1,46 @@
...
+   * @param connected Connected if true
+   * @param conn Connection information
+   */
+  public ConnectionNotification(int listener, boolean connected, ConnectionInfo conn) {
+    this.listener = listener;
+    this.connected = connected;
+    this.conn = conn;
+  }
+
+  NetworkTableInstance inst;
```
AustinShalit: Where is this field set?

PeterJohnson: It's set in the thread in NetworkTableInstance.  This is a little ugly but the most efficient way to do it... I suppose I could pass it to the poll call and have the JNI layer pass it back to the constructor?  That would allow this field to be final.

PeterJohnson: Fixed to pass back through JNI layer to the constructor, so this field is now private final.

PeterJohnson:
@frcjenkins retest this please

PR #208, opened 2017-07-08
GetJavaStackTrace: Provide llvm::StringRef excludeFuncPrefix.

PeterJohnson:
This is a bit cleaner to use than the templated version (many compilers
have issues with templating on static strings, for example).


PeterJohnson:
See https://github.com/wpilibsuite/allwpilib/pull/563

PeterJohnson:
Closing in favor of wpilibsuite/wpiutil#3

PR #209, opened 2017-07-21
Updated Gradle to 4.0.1

calcmogul:


JLLeitschuh:
This looks good.
I would use the full version of gradle instead but that's just me because I do a lot of gradle build development and I like having the docs downloaded too.
Example:
https://github.com/wpilibsuite/allwpilib/blob/master/build.gradle#L67-L70

JLLeitschuh (approved):

PeterJohnson:
There's a major build system update that will be PR'ed soon, so this is OBE.

JLLeitschuh:
"OBE"?

calcmogul:
Obsoleted By Events.

PR #210, opened 2017-08-01
Updates to the new build system

ThadHouse:
Jenkins will fail until I update the scripts there, which I will do in a bit.


build.gradle
```diff
@@ -1,74 +1,192 @@
...
+
+task nativeTestFilesJar(type: Jar) {
+    destinationDir = project.buildDir
+    classifier = "nativeTestFiles"
+
+    project.model {
+        binaries {
+            withType(SharedLibraryBinarySpec) { binary ->
+                if (binary.component.name == 'ntcoreJNI') {
+                    from(binary.sharedLibraryFile) {
```
JLLeitschuh, line 55: Doesn't this cause this to get resolved eagerly at configuration time? Don't you want this to get resolved at task runtime?

config.gradle
```diff
@@ -0,0 +1,226 @@
...
+
+def macCompilerArgs =  ['-std=c++11', '-Wall', '-Wextra', '-Werror', '-pedantic-errors', '-fPIC', '-g',
+                            '-Wno-unused-parameter', '-Wno-missing-field-initializers', '-Wno-unused-private-field']
+def macReleaseCompilerArgs = ['-O2']
+def macDebugCompilerArgs = ['-O0']
+def mac32BitArg = '-m32'
+
+def buildAll = project.hasProperty('buildAll')
+
+def windows64PlatformDetect = { config->
```
JLLeitschuh: Why does this closure take an arg but never use it?

publish.gradle
```diff
@@ -1,132 +1,217 @@
...
+def baseArtifactId = 'ntcore'
+def artifactGroupId = 'edu.wpi.first.ntcore'
 
-project(':native:wpiutil').build.dependsOn outputVersions
-project(':native:ntcore').build.dependsOn outputVersions
-if (project.buildArm) {
-    project(':arm:wpiutil').build.dependsOn outputVersions
-    project(':arm:ntcore').build.dependsOn outputVersions
-}
+def licenseFile = new File("$rootDir/license.txt")
```
JLLeitschuh: Could just use `file("$rootDir/license.txt")`

publish.gradle
```diff
@@ -1,132 +1,217 @@
...
+        into '/'
+    }
 
-                if (project.hasProperty('makeDesktop')) {
-                    artifact nat.jar, {
-                        classifier = 'desktop'
+    model {
+        tasks {
+            it.each {
+                if (it.name.contains('jniHeaders')) {
```
JLLeitschuh: You might also want to check that the type is a `CopySpec`?

publish.gradle
```diff
@@ -1,132 +1,217 @@
...
+    archives sourcesJar
+    archives javadocJar
+    archives cppHeadersZip
+    archives cppSourcesZip
+}
 
-            if (project.hasProperty('makeDesktop')) {
-                artifact nat.ntcoreZip, {
-                    classifier = 'desktop'
+def createComponentZipTasks = { components, name, base, type, project, func ->
```
JLLeitschuh: Maybe turn this into a more strongly typed groovy function instead of using a closure since there are so many inputs? It would help others read it.

src/test/java/TestClass.java
```diff
@@ -0,0 +1,17 @@
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import static org.junit.Assert.assertEquals;
+
+public class TestClass {
+    @Test
+    public void generalTestCase() {
+        assertEquals(true, true);
```
JLLeitschuh: What is this testing? Perhaps a bit more documentation if you are actually expecting something here.

src/test/java/TestClass.java
```diff
@@ -0,0 +1,17 @@
+import org.junit.BeforeClass;
```
JLLeitschuh: This seems to be in the default package. This is not really done even in test systems. You should put this test in its own package.

build.gradle
ThadHouse: It does not. The native builds handle eagerly configuring this. It does get resolved at runtime.

config.gradle
ThadHouse: I always forget in gradle you can ignore closure arguments. There are other cases where that config is used, but I can remove it from here.

publish.gradle
ThadHouse: Good idea. Something similar to this is used in most of the repo's, so thats a good idea.

src/test/java/TestClass.java
ThadHouse: Ok I will move it.

src/test/java/TestClass.java
ThadHouse: This was more to verify that tests were actually running, and the test below is to make sure the JNI linking actually works. I'll remove the general one, and move the JNI one to it's own package.

publish.gradle
ThadHouse: Yeah probably.

ThadHouse: Actually the task it is looking at is not a copy task, so that doesn't work. This is one of the few cases where it is just hard to get something out of the model.

publish.gradle
ThadHouse: I'm going to pass on this, because I use a few script globals, and switching to a function breaks that, with the workarounds looking super hacky

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTablesJNI.java
```diff
@@ -67,12 +55,6 @@ else if (System.getProperty("os.name").startsWith("Mac"))
     }
   }
 
-  public static final int NT_NET_MODE_NONE = 0x00;
-  public static final int NT_NET_MODE_SERVER = 0x01;
-  public static final int NT_NET_MODE_CLIENT = 0x02;
-  public static final int NT_NET_MODE_STARTING = 0x04;
-  public static final int NT_NET_MODE_FAILURE = 0x08;
-
```
PeterJohnson: This seems like an unintentional revert.  Are you synced up to the latest master?

src/main/java/edu/wpi/first/wpilibj/networktables/NetworkTablesJNI.java
```diff
@@ -154,7 +136,6 @@ else if (System.getProperty("os.name").startsWith("Mac"))
   // public static native byte[] getRpcResultNonblocking(int callUid) throws RpcNoResponseException;
 
   public static native void setNetworkIdentity(String name);
-  public static native int getNetworkMode();
```
PeterJohnson: Ditto

PeterJohnson (changes requested):

publish.gradle
```diff
@@ -1,132 +1,222 @@
...
-}
+def outputsFolder = file("$buildDir/outputs")
 
-def utilFile = file("$buildDir/wpiutil.txt")
-def ntcoreFile = file("$buildDir/ntcore.txt")
+def versionFile = file("$outputsFolder/version.txt")
 
 task outputVersions() {
-    description = 'Prints the versions of ntcore and wpiutil to a file for use by the downstream packaging project'
+    description = 'Prints the versions of wpiutil to a file for use by the downstream packaging project'
```
PeterJohnson: s/wpiutil/ntcore/

PeterJohnson (changes requested):

build.gradle
```diff
@@ -1,74 +1,196 @@
...
+            }
+        }
+    }
+}
 
 apply from: 'publish.gradle'
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '3.3'
+    gradleVersion = '4.0.1'
```
AustinShalit: Should we bump to `4.0.2`?

JLLeitschuh: Yes

publish.gradle
JLLeitschuh: What type is it then?

ThadHouse: https://github.com/wpilibsuite/native-utils/blob/master/src/main/groovy/edu/wpi/first/nativeutils/rules/JNIConfigRules.groovy#L40
It's just a Task, it doesn't inherit from anything.

publish.gradle
ThadHouse: Fixed

build.gradle
ThadHouse: fixed

publish.gradle
JLLeitschuh: Reading through that code, just as an aside, its much better practice to create types for your tasks. Doing so allows you to use the annotations `@Input`, `@InputDirectory`, `@Output`, `@OutputDirectory` ect... That allow your build to do things better with task avoidance (speeding up the build).
Additionally, doing so would allow you to take advantage of the new gradle build cache where you can share the output of tasks on a central server (or even on your personal machine) to avoid running tasks that you've already run before.

Additionally, I worry that the lack of any sort of type system here will make this code unmaintainable by anyone in the future. Without a solid set of tests to prove that what the build is creating is correct it's much harder for new developers to dive in without the fear of breaking something.

ThadHouse: How are the annotations different from just declaring Inputs and Outputs on the task, which I already do. Also, until all the native stuff supports the build cache (which none of it does currently), I don't think its necessary to take the extra work. I do agree that adding a custom type for this might be a good idea however. I just wish the team had commented on this weeks ago when I asked for help and suggestions rather then waiting until we were wanting to merge it.

publish.gradle
```diff
@@ -1,132 +1,222 @@
...
-                if (project.hasProperty('makeDesktop')) {
-                    artifact nat.jar, {
-                        classifier = 'desktop'
+    model {
+        tasks {
+            it.each {
+                if (it.name.contains('jniHeaders')) {
+                    from (it.outputs.files) {
+                        into '/'
                     }
```
JLLeitschuh: When does this execute? It looks like it runs during task configuration time, not at task run time? Is that what you want?

ThadHouse: This does execute at run time. Otherwise copy and zip tasks would be absolutely useless.

JLLeitschuh: Can you explain how that's possible? It seems like the receiver of the call to `from` is `project` not the task. As you said above, the task is of type `DefaultTask` which doesn't have a `from` method on it.

If the receiver of the `from` call is `project` then it would be executed immediately.

What is the receiver of the `from` call?

ThadHouse: The cppSourcesZip. GitHub is just splitting the lines wierdly.

JLLeitschuh: Oooooh. That makes so much more sense.

publish.gradle
ThadHouse: @JLLeitschuh I created custom types for my tasks. I didn't add anything to them, since the default task worked just fine, but it does let me do type checking.

PeterJohnson (approved):

PR #213, opened 2017-08-04
Remove deprecated throwing get functions.

PeterJohnson:


AustinShalit (commented):
We should wait to merge this until SmartDashboard is updated to not use these methods.  I should able to update it tomorrow.

PR #215, opened 2017-08-04
Fix ntcoreExe build model.

PeterJohnson:


ThadHouse:
Do you want this to also include the source directory of ntcore for includes?

PeterJohnson:
No.  At least how I think dev is used, unlike test, it should only use the public API?

PR #218, opened 2017-08-08
Adds a java devExe

ThadHouse:


build.gradle
```diff
@@ -202,6 +202,22 @@ model {
...
+project(':devApp') {
+    apply plugin: 'java'
+    apply plugin: 'application'
+
+    dependencies {
+        compile project(':')
+        runtime files(project(':').nativeTestFilesJar.archivePath)
+    }
+
+    sourceSets.main.java.srcDirs = [ "${rootDir}/src/dev/java" ]
```
JLLeitschuh: Why don't you just follow the gradle standard and create a new directory called `devApp` and put the source code in there?

What you are doing here seems hacky and goes against the gradle mentality.

You would normally see another source set in a project's `src directory, not an entirely different project's source code.

ThadHouse: Because we decided we wanted consistency between c++ and Java, so that's the folder layout we wanted. It's just a test app for developers anyway. Other people didn't like the subprojects idea for CPP.

ThadHouse: I wish gradle supported Java as components like cpp, not only subprojects. Its a much cleaner layout imo.

JLLeitschuh: It does, you can create additional source sets.

ThadHouse: So that will create a separate jar?

ThadHouse: So like I can make the jar use the main source set, and run use main and dev?

JLLeitschuh: You can create a separate jar from a specific source set. You need to define a JAR task that pulls from that source set. You may be able to configure the application plugin to use that source set instead of the default `src` one.

This is why I say it's hacky. Given this is a gradle project you should work with it and not fight it like this. But you also shouldn't put another project's source code under its parent's `src` directory.

ThadHouse: Id prefer uniformity over perfect, since it's either going to make the c++ devs unhappy or the java devs unhappy. And for these libs, since CPP is their main component I think I'd prefer keeping with cpp. I get what you mean by it being hacky, but we are just not doing something supported anyway. I would love if we had a build system that was perfect, but nothing is. Making the repo look like crap with code all over the place just to do a Java test app the Gradle way seems dumb to me.

ThadHouse: I'll try and get rid of the subprojects, as I do hate the idea of it. But I'm ok with a jar hack if it keeps the repo clean.

ThadHouse:
@JLLeitschuh I found a much cleaner way of running this. It uses a JavaExec task rather then the application plugin, since this doesn't need the distribution features anyway (or a jar for that matter, which it skips as well.)

AustinShalit:
Why do we need this?  Is it to check that Java can run properly?

ThadHouse:
Its a built in test exe so you can test Java code without needing another project. I added the same thing to the C++ side, and it helps a ton.

JLLeitschuh:
Looks much better!

ThadHouse:
Agreed. This looks much cleaner, and more configurable. That gradle application plugin is has no configuration capabilities, which doesn't help..

JLLeitschuh:
Just curious, how is this dev enviroment any different from simply running the JunitTests?

You could check to make sure that this works just by running `./gradlew check` right?

Shouldn't any testing you are doing in the Java world get captured in a Junit test?

ThadHouse:
This is for like API development. Since this is a library, there isn't anything to run. Sometimes I've just wanted a main() to be a sandbox for testing random things and developing new ideas. Like things that are not worth unit tests, but setting up a whole new project is just a pain. Technically you could just use the unit tests, but right tool for the right job, and IMO debugging an app is easier then debugging some randomly named unit test that is being ran with reflection. Its more of a development exe, not test exe.

JLLeitschuh:
API development soulds like the perfect time to apply [Test Driven Development](https://en.wikipedia.org/wiki/Test-driven_development).

Are you debugging from the command line or with an IDE? With any IDE debugging a test method should be painless. You can run one test in intellij and it will automatically attach a debugger allowing you to set breakpoints, ect.

Plus, any testing you do involving the API, if it becomes production code, will already be tested.

ThadHouse:
I can never figure out how to get any Java IDE working, so I'm doing all of mine from the command line. And I don't want to have to wait for all the other tests to run. From the cpp side, a sandbox exe is usually a good thing to have, and for java dev I know most java devs work with ide's, but some of us just have more trouble with those and a simple running exe is better.

ThadHouse:
IMO, unit tests are not the end all be all of all issues. Some things just need to be figured out manually, especially JNI issues, and hacking a unit test to end up being a main just seems like the wrong tool for the job. For instance, It took me 40 minutes the other day to figure out how to run unit tests with custom JVM args in gradle, since its not well documented, whereas with the JavaExec task its right in the task definition.

JLLeitschuh:
If you want help getting IntelliJ setup I can help. Legit, IntelliJ imported the project and I was running that one Java test in seconds.

I ran `./gradlew build -PskipArm` and then ran the test in intellij and it passed without any pain.

PeterJohnson (approved):

PR #235, opened 2017-09-26
Depend on wpiutil 3.+ rather than just +.

PeterJohnson:
Also update to wpilib-version-plugin 2.0.


PeterJohnson:
Depends on wpilibsuite/wpiutil#39.

JLLeitschuh:
I'm going to recommend being a bit more exact so that you can run a git bisect and find where bugs come from a bit more exactly.

PeterJohnson:
A bit more exact meaning what?  I do NOT want to specify exact versions.

PR #239, opened 2017-10-01
Don't actually remove deleted entries from m_entries.

### LINKED ISSUE #238: entry instance handles are completely invalid when a name is deleted
I haven't actually tried it yet, but it seems fairly clear that if you do (referring to the storage object only):

* SetEntry
* Get entry handle
* DeleteEntry by name
* SetEntry

The handle retrieved is no longer valid because the local_id isn't associated with a valid entry (and actually... the entry is a shared_ptr... so I wonder if it would think it's still valid?), so the second set creates a new entry with a new local_id. This occurs because entries are removed from m_entries upon deletion (Storage::DeleteEntryImpl), and that's the only way to associate an entry with a local_id.

PeterJohnson:
Doing so invalidates entry instance handles.

Fixes #238.


virtuald:
Seems like it should do the trick.

PR #241, opened 2017-10-02
Fix handling of deleted values in several places.

PeterJohnson:
Now that the m_entries entry is kept, other places can't assume that
a value exists simply because the entry exists.


virtuald (approved):
LGTM.

PR #243, opened 2017-10-02
Added sample Gradle config to use pre-built ntcore

flybotix:
Hello,
I'd like to contribute this, and potentially further, updates to the documentation of WPILib in an effort to make using it a little easier.  This took a bit of trial-and-error to get right, but now that I have it figured out I hope others find it useful.

Regards,
Jesse Knight (JesseK on CD)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,21 @@
 
 ntcore is the reimplementation of the NetworkTables protocol for both Java and C++, communicating with the former by way of a JNI interface. ntcore implements v3 of the NetworkTables spec.
 
+## Using ntcore Artifacts
+To use ntcore without downloading the entire WPILib suite, here is a sample Gradle configuration.  Note that while this will download the appropriate runtime files (e.g. ntcore.dll) it does not cover the necessary steps to add the file to the runtime path. In other words, runtime dependencies must be configured manually if using ntcore as a standalone library.
+```bash
+repositories {
+  maven {
+    url 'http://first.wpi.edu/FRC/roborio/maven/release'
+  } // WPILib
+}
+
+dependencies {
+  compile group: 'edu.wpi.first.wpilib.networktables.java', name:'NetworkTables', version:'3.1.7', classifier : 'desktop'
+  compile group: 'edu.wpi.first.wpilib.networktables.java', name:'NetworkTables', version:'3.1.7', classifier : 'sources'
+}
+```
+
 ## Build Requirements
 To build ntcore, a few requirements must be met:
 


ThadHouse:
For the 2018 season, the artifact location has changed, in addition to requiring more artifacts to run completely. Once we have a release, you can either update this PR, or I can just write the equivalent with the new locations.

flybotix:
@ThadHouse Either is fine.  I'll keep using this on my local project for now, and update it with the 2018 release when it's available.

README.md
```diff
@@ -6,6 +6,21 @@
 
 ntcore is the reimplementation of the NetworkTables protocol for both Java and C++, communicating with the former by way of a JNI interface. ntcore implements v3 of the NetworkTables spec.
 
+## Using ntcore Artifacts
+To use ntcore without downloading the entire WPILib suite, here is a sample Gradle configuration.  Note that while this will download the appropriate runtime files (e.g. ntcore.dll) it does not cover the necessary steps to add the file to the runtime path. In other words, runtime dependencies must be configured manually if using ntcore as a standalone library.
+```bash
```
JLLeitschuh, line 11: I usually use `groovy` or `gradle` for the syntax highlighting hint.

luisnaranjo733:
Did the paths end up changing? Just wondering if this is still usable so I can recommend it to the students I am mentoring. This would be invaluable.

Our team is trying to build this on a 64 bit raspberry pi, which doesn't have the gcc-multilib 64 bit dependency available, so we can't build it for ourselves.

ThadHouse:
We don't have builds for the pi, and especially for 64 bit. Does your system not have gcc at all? If it does, and it has cmake I have a cmake build system I would recommend using.

luisnaranjo733:
We do have GCC installed, but we weren't able to install the gcc-multilib library via apt-get.

We are trying to install cmake now. How do we use this cmake build system?

flybotix:
The paths for 3.1.7 haven't changed.  However, I would like the info for v4+.  We haven't gotten around to implementing ntcore outside in our display code yet, but with what we plan to do in Week 4-5, we will.  Are there changes?

ThadHouse:
All the new files are located in the following directory.

http://first.wpi.edu/FRC/roborio/maven/release/edu/wpi/first/

ntcore and wpiutil are located in the ntcore and wpiutil folders respectively. Everything is now a lot better named and placed.

As for the cmake build, it is located here

https://github.com/thadhouse/CmakeWpilib

it used submodules, so you need to clone the repo recursively. It also builds cscore and java, which you might not need. cscore can be disabled with `-DWITHOUT_CSCORE=ON` and java can be disabled with `-DWITHOUT_JAVA=ON`

luisnaranjo733 (commented):
How do we add the files to our runtime path?

PR #249, opened 2017-10-14
Removes specific raspbian, armv7 and armv8 cross builds

ThadHouse:
Matches wpiutil #41

### DIFF
--- a/config.gradle
+++ b/config.gradle
@@ -175,10 +175,17 @@ if (!project.hasProperty('onlyAthena')) {
                 compilerFamily = 'Clang'
                 detectPlatform = mac64PlatformDetect
             }
-            raspbian(CrossBuildConfig) {
-                architecture = 'raspbian'
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


config.gradle
```diff
@@ -175,34 +175,17 @@ if (!project.hasProperty('onlyAthena')) {
                 compilerFamily = 'Clang'
                 detectPlatform = mac64PlatformDetect
             }
-            raspbian(CrossBuildConfig) {
```
JLLeitschuh, line 178: Does this not work? Why no build for the PI anymore?

ThadHouse: It was running into both maintainability issues and other extraction issues that were difficult to solve. The solution for now is to make all of them build as a `nativearm` classifier. It makes the custom user case easier, and the custom cross case easier as well. And it was never really tested and kind of a pain to work with as well.

PeterJohnson (approved):

PR #252, opened 2017-11-09
Adds header task

ThadHouse:


JLLeitschuh:
Can you explain what this is for? I've seen it elsewhere and I don't understand what it's supposed to do.

ThadHouse:
I'm building a vscode plugin that grabs all the header dirs and generates the include paths so intellisense works. Can also be used for building other plugins as well.

PR #256, opened 2017-11-19
NetworkTable: Add key utility functions.

PeterJohnson:
- BasenameKey
 - NormalizeKey
 - GetHierarchy


src/main/java/edu/wpi/first/networktables/NetworkTable.java
```diff
@@ -28,6 +30,92 @@
   private final String pathWithSep;
   private final NetworkTableInstance inst;
 
+  /**
+   * Gets the "base name" of a key. For example, "/foo/bar" becomes "bar".
+   * @param key key
+   * @return base name
+   */
+  public static String basenameKey(String key) {
```
SamCarlberg: Is it worth mentioning that a key with a trailing slash will output an empty string `""`?

SamCarlberg (dismissed):

PR #259, opened 2017-11-23
Use Twine instead of StringRef where appropriate.

PeterJohnson:
Deprecated interfaces were not updated.


JLLeitschuh:
What is `Twine`?

PeterJohnson:
http://llvm.org/docs/ProgrammersManual.html#the-twine-class

PR #263, opened 2017-12-21
Adds compile task

ThadHouse:
Reenables travis builds, just without linking.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -11,7 +11,7 @@ buildscript {
         }
     }
     dependencies {
-        classpath 'gradle.plugin.edu.wpi.first:native-utils:1.3.0'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -240,6 +240,15 @@ model {
                 }
             }
         }
+        compileCpp(Task) {
+            $.binaries.each { binary ->
```
JLLeitschuh, line 244: What's the `$` mean in groovy?

ThadHouse:
It uses other sources of the Gradle model setup as an input. So this gives you access to all the model binaries.

build.gradle
333fred: `$` is a gradle-specific symbol, it means model root.

JLLeitschuh: Thanks!

PR #270, opened 2018-02-28
Fix two bugs in client synchronization.

PeterJohnson:
Both could occur if a client and server write to the same key and the server
disconnects/reconnects (or restarts).

Bug 1: the client did not properly update the sequence number in this case,
so later server updates could be ignored until the sequence number wrapped.

Bug 2: the client did not properly set the id and sequence number for the
update message back to the server, so the server would ignore the message.


bhjelstrom:
Fixes Limelight sync issue

virtuald:
I know this needs to get out ASAP, but it would be good to put in a unit test that covers this.

virtuald (approved):

PR #272, opened 2018-03-15
Malloc checks

XavilPergis:
I just added some checks for null pointers where they are appropriate.


XavilPergis:
Whoops, I forgot some nots!

PeterJohnson:
Can you change this to call std::terminate (http://en.cppreference.com/w/cpp/error/terminate) instead of exit?

JLLeitschuh:
Include a message please.
Another potential, possibly simpler solution, is to create a common wrapper around malloc that will automatically call `terminate` so that this logic doesn't need to be added at every call site?

PeterJohnson:
Superseded by wpilibsuite/allwpilib#1023.

PR #274, opened 2018-03-16
Fix ability to compile with gcc

rjbell4:
With these changes, I am able to compile a small C program with gcc

### DIFF
--- a/src/main/native/include/ntcore_c.h
+++ b/src/main/native/include/ntcore_c.h
@@ -10,7 +10,12 @@
 
 #include <stdint.h>
 
+#ifdef __cplusplus
 #include <cstddef>
+#else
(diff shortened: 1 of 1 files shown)


PeterJohnson:
While it works, the way the ifdef is done doesn't feel very idiomatic.  I'd prefer one ifdef for the headers and a separate one for the extern "C".

rjbell4:
Sure, I was just trying to reduce the `#ifdef`s, since they were on the same condition.  Sorry for the day; I've updated the pull request now.

ThadHouse (approved):

PeterJohnson:
Moved to wpilibsuite/allwpilib#1007 now that ntcore is merged into allwpilib.

